PR #3, opened 2014-01-16
Added Project Naming Strategy check to existing folder names

christ66:
Currently the Project Naming Strategy only works for Jobs when force existing is true. This pull request enables the Project Naming Strategy when forced enabled is enabled.


bitwiseman:
Ancient.  Please comment if you want to pick up the PR.

PR #10, opened 2015-03-21
[JENKINS-23879] Add a move command CLI operation

christ66:
WIP because I need to add a few more tests.

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/relocate/MoveCommand.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/relocate/MoveCommand.java
@@ -1,0 +1,62 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2015 CloudBees., Steven Christou
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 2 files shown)


daniel-beck:
I'd like to see a test that shows that builds of the moved project are available in the new location. I have serious doubts this happens.

src/main/java/com/cloudbees/hudson/plugins/folder/relocate/MoveCommand.java
```diff
@@ -0,0 +1,62 @@
...
+    @Override
+    protected int run() throws Exception {
+        Jenkins jenkins = Jenkins.getInstance();
+        if (jenkins.getItemByFullName(dst) != null) {
+            stderr.println("Job " + dst + " already exists.");
+            return -1;
+        }
+
+        jenkins.copy(src, dst).save();
+        jenkins.remove(src);
```
jglick: This is not even close to correct. You must use `RelocationHandler`. See `RelocationAction.doMove`.

jglick: (The “close to correct” version would call `Items.move`, as `StandardHandler` does. That would take care of moving build history, for example; the tests for that are already in core. It would not handle additional features injected by other `RelocationHandler`s such as are added by `cloudbees-folders-plus`.)

bitwiseman:
Stale. Closing for now.  Please feel free to reopen if appropriate.

PR #11, opened 2015-04-05
[JENKINS-27735] Disabling a folder

alecharp:
When folders are used to contain all jobs for a release or a feature, been able to disable all jobs in one click is a true time-saving feature.

Issue: [JENKINS-27735](https://issues.jenkins-ci.org/browse/JENKINS-27735)

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/buildable/DisableAction.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/buildable/DisableAction.java
@@ -28,6 +28,7 @@ import com.cloudbees.hudson.plugins.folder.Folder;
 import com.cloudbees.hudson.plugins.folder.Messages;
 import com.google.common.base.Predicate;
 import com.google.common.collect.Collections2;
+import com.google.common.collect.Lists;
 import com.google.common.collect.Ordering;
 import hudson.Extension;
 import hudson.model.AbstractProject;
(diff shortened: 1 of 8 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/buildable/DisableAction.java
```diff
@@ -0,0 +1,125 @@
...
+     */
+    @RequirePOST
+    public HttpResponse doDisable(StaplerRequest req) throws IOException {
+        LOGGER.finest("Trying to disable all jobs in " + item.getName());
+        for (Job job : getPossibleJobs()) {
+            try {
+                ((AbstractProject) job).disable();
+                LOGGER.finest("Set " + job.getName() + " as not buildable");
+            } catch (ClassCastException e) {
+                LOGGER.info("Cannot disable " + job.getName());
```
daniel-beck: CCE is not something a user cares about, it just means the API for doing this is weird. This should be handled internally, and displayed to the user as "The following cannot be disabled".

src/main/resources/com/cloudbees/hudson/plugins/folder/buildable/DisableAction/index.jelly
```diff
@@ -0,0 +1,22 @@
...
+    <st:include page="sidepanel.jelly" it="${it.item}"/>
+    <l:main-panel>
+      <h1>
+        ${%Disable.title(it.item.getDisplayName())}
+      </h1>
+
+      <form action="disable" method="POST">
+      ${%confirmationMessage}
+        <ul>
+          <j:forEach var="it" items="${it.getJobs()}">
```
daniel-beck: This probably shows fewer jobs than will get disabled (nested folders?). In general, this should probably not use a different source for the list of items than the actual action.

Plus, false positives in case of the already mentioned CCE could be caught and listed separately.

alecharp: the `getAllJobs()` method on a folder return all jobs, even in nested folders. Using the same source as the action shows only the jobs that will be changed.

I agree about the CCE. I'm looking on how to give the user the list of missed jobs.

Matt029847294:
Any news on this one yet? Feature would be really useful to cut down on an overcrowded view.

bitwiseman:
Stale. Closing for now.  Please feel free to reopen if appropriate.

PR #12, opened 2015-04-11
Allow configuration of view permissions in folders

joshua-jandyco:
When configuring a folder, the view group of permissions will now be configurable.  You may now configure whether users can create/delete/configure or read views within a folder.

Maybe there is more work than I know of or some other constraints that are not taking into account as I'm not sure why this wasn't included to begin with.  Please let me know if so.


jglick:
Subsumed by #13.

PR #18, opened 2015-09-17
Add support for the `TopLevelItemDescriptor.isApplicableIn(ItemGroup)` method in 1.607+

stephenc:
Don't want to bump the core version of Jenkins just yet, so using reflection for now.
The use of reflection should not be a performance risk as the New Item page is not rendered
heavily, and in any case reflection gets inlined well by the JVM

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/Folder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/Folder.java
@@ -806,7 +806,7 @@ public class Folder extends AbstractItem
         }
         // TODO Jenkins 1.607+ use method directly
         try {
-            return Boolean.TRUE.equals(tid.getClass().getMethod("isApplicableIn", ItemGroup.class).invoke(this));
+            return Boolean.TRUE.equals(tid.getClass().getMethod("isApplicableIn", ItemGroup.class).invoke(tid, this));
         } catch (NoSuchMethodException e) {
             // ignore
         } catch (InvocationTargetException e) {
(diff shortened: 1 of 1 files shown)


recena:
:bee:, Why don't we want to update the Jenkins baseline?

oleg-nenashev:
:bee: I would not vote for the dependency update, because we still need to support 1-year-old LTS lines.

stephenc:
There may be a bug in the code... running by hand tests now (by hand tests needed as we cannot up the dependency on core until November)

stephenc:
@recena @oleg-nenashev care to go again... my bad for not finishing manual testing before creating the PR

oleg-nenashev:
re :bee:

recena:
@stephenc Well catch it out :bee:

src/main/java/com/cloudbees/hudson/plugins/folder/Folder.java
```diff
@@ -803,6 +804,16 @@ public boolean isAllowedChildDescriptor(TopLevelItemDescriptor tid) {
...
         }
+        // TODO Jenkins 1.607+ use method directly
+        try {
+            return Boolean.TRUE.equals(tid.getClass().getMethod("isApplicableIn", ItemGroup.class).invoke(tid, this));
+        } catch (NoSuchMethodException e) {
+            // ignore
+        } catch (InvocationTargetException e) {
+            // ignore
+        } catch (IllegalAccessException e) {
+            // ignore
```
jglick: The second two exceptions would be genuine problems that should be rethrown as `AssertionError`, not ignored.

jglick:
:ant: on using reflection when the API is in an LTS line. If at some point we want to publish unrelated changes for users of other lines, we can cut releases from a sustaining branch.

jglick:
> we cannot up the dependency on core until November

We can increase the dependency whenever we find it useful to do so. It is true that as of Oct 30 CloudBees will stop supporting Jenkins installations older than 1.609.3, but so what? If there is a need to backport fixes for customers in the next three weeks, we can do that easily, and for that matter if anyone else really needs critical fixes for much older cores we can cherry-pick them to the existing `stable` branch.

PR #17, opened 2015-09-10
Add support for dynamic icon based on the state of children

duhanebel:
The folder icon is currently static. This commit add a new icon type
that allows the folder state icon to reflect the children status:
- color based on worst child build status: red, yellow, "blue" (green with
  GreenBalls plugin), etc.
- animated when 1+ child build is running.

Fixes JENKINS-20715

PR #19, opened 2015-09-30
Set folder object for icons that already exist

0x4a616e:
When changing the folder icon class using the jenkins dsl plugin or by manually editing the config file, the `FolderIcon` seems not to get the `Folder` object because the icon it already exists when `Folder.init` is called. Setting the folder object for existing icons seems to solve this issue.


0x4a616e:
Sorry for breaking the build, but I can hardly imagine how this could be cause by my commit... For me it looks more like some kind of internal build error probably related to findbugs. I'm also not able to reproduce this locally :confused:

jglick:
Looks right, though a test would be useful. BTW `StockFolderIcon` does not use this backlink, so how did you find this?

0x4a616e:
You're right, I'll see if I can add a test! I found this while playing around with the job-dsl-plugin and cloudbees-folder-plugin. When I tried to create an additional plugin to set the folder icons according to the enclosed project's build states, I noticed that `setFolder` was never called. I think it is called in `doConfigSubmit` only, which i believe is not called when modifying the configuration file directly using the job-dsl-plugin.

jglick:
Needs merge with `master`; the relevant code is now in `AbstractFolder` after #20 (and the method is now `setOwner`). I do suspect the patch is still needed, but only for `AggregateFolderIcon` in the proprietary `cloudbees-folders-plus`, and for the proposed icon in #17.

0x4a616e:
Merged with master, changes could be merged back again.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -190,6 +190,8 @@ protected void init() {
         }
         if (icon == null) {
             icon = new StockFolderIcon();
```
stephenc: I know that `StockFolderIcon` does not use the `owner` field... but that does not mean that somebody won't change it in the future... it would seem to be better to have this always call `icon.setOwner(this);` even if *currently* that is a no-op for `StockFolderIcon`

PR #20, opened 2015-10-07
ComputedFolder

jglick:
A new API for folders whose contents are determined algorithmically rather than by direct user modification.

- [X] baseline update
- [x] add `ComputedFolder`
- [x] make `ComputedFolder` extend `AbstractFolder`
- [x] redesign API
- [x] define `OrphanedItemStrategy`
- [x] `PeriodicFolderTrigger`
- [x] `Cron`
- [x] `List<Trigger> triggers`
- [x] `ThrottleComputationQueueTaskDispatcher`
- [x] narrow type of `FolderComputation.getParent`

@reviewbybees esp. @stephenc


tfennelly:
Is all the new code (`ComputedFolder` etc) coming from somewhere else? Are there tests?

jglick:
Yes and yes.

stephenc:
looks :bee: so far

rsandell:
:bee:

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -0,0 +1,746 @@
...
+        Jenkins.checkGoodName(newName);
+
+        String blocker = renameBlocker();
+        if (blocker != null) {
+            rsp.sendRedirect("rename?newName=" + URLEncoder.encode(newName, "UTF-8") + "&blocker=" + URLEncoder.encode(blocker, "UT-8"));
+            return;
+        }
+
+        renameTo(newName);
+        rsp.sendRedirect2("../" + newName);
```
amuniz: `../`? Why not `newName` (realtive) directly?

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderProperty.java
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Property potentially applicable to any {@link AbstractFolder}.
+ * @since FIXME
+ */
+public abstract class AbstractFolderProperty<C extends AbstractFolder<?>> extends AbstractDescribableImpl<AbstractFolderProperty<?>> implements ExtensionPoint, ReconfigurableDescribable<AbstractFolderProperty<?>> {
+
+    /**
+     * The {@link AbstractFolder} object that owns this property.
+     * This value will be set by the Hudson code.
```
amuniz: Jenkins, no?

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -0,0 +1,746 @@
...
+
+    private transient /*almost final*/ ViewGroupMixIn viewGroupMixIn;
+
+    private DescribableList<FolderHealthMetric,FolderHealthMetricDescriptor> healthMetrics;
+
+    private FolderIcon icon;
+
+    /** Subclasses should also call {@link #init}. */
+    protected AbstractFolder(ItemGroup parent, String name) {
+        super(parent, name);
```
amuniz: Why not calling `init` here too?

src/main/java/com/cloudbees/hudson/plugins/folder/Folder.java
```diff
@@ -163,235 +101,58 @@
...
-     */
-    private volatile String primaryView;
-
-    private transient /*almost final*/ ViewGroupMixIn viewGroupMixIn;
-
-    private DescribableList<FolderHealthMetric,FolderHealthMetricDescriptor> healthMetrics;
-
     public Folder(ItemGroup parent, String name) {
         super(parent, name);
         init();
```
amuniz: It could be moved to `AbstractFolder` since it must be called by all subclasses.

amuniz: Oh, it's overridden here, never mind.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
```diff
@@ -0,0 +1,439 @@
...
+
+    public boolean isBuildable() {
+        return true;
+    }
+
+    @RequirePOST
+    public HttpResponse doBuild(@QueryParameter TimeDuration delay) {
+        checkPermission(BUILD);
+        if (!isBuildable()) {
+            throw HttpResponses.error(SC_INTERNAL_SERVER_ERROR, new IOException(getFullName() + " cannot be indexed"));
```
amuniz: ` cannot be computed` would be more coherent.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
```diff
@@ -0,0 +1,247 @@
...
+import java.util.List;
+import static java.util.logging.Level.FINE;
+import static java.util.logging.Level.FINER;
+import java.util.logging.Logger;
+import org.apache.commons.lang.StringUtils;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+/**
+ * The default {@link OrphanedItemStrategy}.
+ * Trims dead items by the # of days or the # of jobs much like {@link LogRotator}.
```
amuniz: `# of jobs`? I don't see code doing anything based on number of jobs. Perhaps I'm missing something.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
```diff
@@ -0,0 +1,247 @@
...
+
+    /**
+     * Our logger
+     */
+    private static final Logger LOGGER = Logger.getLogger(DefaultOrphanedItemStrategy.class.getName());
+
+    /**
+     * Should old branches be removed at all
+     */
+    private final boolean pruneDeadBranches;
```
amuniz: Why the term `branches` is used? They are `item`s AFAIU

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -0,0 +1,746 @@
...
+                initViews(views);
+            } catch (IOException e) {
+                LOGGER.log(Level.WARNING, "Failed to set up the initial view", e);
+            }
+        }
+        if (viewsTabBar == null) {
+            viewsTabBar = new DefaultViewsTabBar();
+        }
+        if (primaryView == null) {
+            primaryView = views.get(0).getViewName();
```
tfennelly: I see earlier that `initViews` can fail and you fall through. Don't you need to check again if the `views` list `isEmpty`?

src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderCron.java
```diff
@@ -0,0 +1,140 @@
...
+    private final Field tabsField;
+
+    /**
+     * Constructor.
+     *
+     * @throws NoSuchFieldException
+     */
+    // instantiated by Jenkins
+    @SuppressWarnings("unused")
+    public FolderCron() throws NoSuchFieldException {
```
amuniz: Why not catching and ignore it? It is never going to be thrown (the `tabs` field was introduced in 1.312)

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
tfennelly: I see later (in the `initViews` impl) that `AllView ` gets added and is probably guaranteed to create without an error. It's probably the `save` that fails?

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ThrottleComputationQueueTaskDispatcher.java
```diff
@@ -0,0 +1,98 @@
...
+ */
+@SuppressWarnings("unused") // instantiated by Jenkins
+@Extension
+public class ThrottleComputationQueueTaskDispatcher extends QueueTaskDispatcher {
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public CauseOfBlockage canRun(Queue.Item item) {
+        if (item.task instanceof ComputedFolder) {
```
amuniz: Could a `ComputedFolder` be concurrently indexed twice? I don't think so. Perhaps a check to avoid that would be in place.

amuniz:
:bee: (as far as I get it)

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -0,0 +1,746 @@
...
+
+    @Override
+    public synchronized void doSubmitDescription(StaplerRequest req, StaplerResponse rsp) throws IOException, ServletException {
+        getPrimaryView().doSubmitDescription(req, rsp);
+    }
+
+    public void doConfigSubmit(StaplerRequest req, StaplerResponse rsp) throws IOException, ServletException, Descriptor.FormException {
+        checkPermission(CONFIGURE);
+
+        req.setCharacterEncoding("UTF-8");
```
tfennelly: I think a comment here would be useful as this looks a bit strange (setting the char encoding of the request).

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -0,0 +1,746 @@
...
+    public void renameTo(String newName) throws IOException {
+        super.renameTo(newName);
+    }
+
+    @Override
+    public synchronized void doSubmitDescription(StaplerRequest req, StaplerResponse rsp) throws IOException, ServletException {
+        getPrimaryView().doSubmitDescription(req, rsp);
+    }
+
+    public void doConfigSubmit(StaplerRequest req, StaplerResponse rsp) throws IOException, ServletException, Descriptor.FormException {
```
tfennelly: `@Restricted(DoNotUse.class)` ?

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -0,0 +1,746 @@
...
+            rsp.sendRedirect("rename?newName=" + URLEncoder.encode(newName, "UTF-8"));
+        } else {
+            if (namingStrategy.isForceExistingJobs()) {
+                namingStrategy.checkName(name);
+            }
+            FormApply.success(".").generateResponse(req, rsp, this);
+        }
+    }
+
+    protected void submit(StaplerRequest req, StaplerResponse rsp) throws IOException, ServletException, Descriptor.FormException {}
```
tfennelly: `@Restricted(DoNotUse.class)` ?

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -0,0 +1,746 @@
...
+
+    // TODO boilerplate like this should not be necessary: JENKINS-22936
+    @RequirePOST
+    public void doDoRename(StaplerRequest req, StaplerResponse rsp) throws IOException, ServletException {
```
tfennelly: `@Restricted(DoNotUse.class)` ?

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ChildObserver.java
```diff
@@ -0,0 +1,61 @@
...
+     */
+    public abstract boolean mayCreate(String name);
+
+    /**
+     * Notify the observer that you did create a new child.
+     * @param child a newly constructed child item; do not call {@link Item#onCreatedFromScratch} and try to avoid calls to {@link Item#save}
+     */
+    public abstract void created(I child);
+
+}
```
tfennelly: The `shouldUpdate` and `mayCreate` methods don't really sound like event methods (events that an "observer" might be interested in). Should they be somewhere else, or should `ChildObserver` be called something else?

tfennelly:
:bee: insofar as I get it (it's a monster).

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
stephenc: AIUI there are two paths. Either the new instance is created before XStream injects the field values from the persisted form and then `onLoad` gets called, or the new instance is created and then `onCreatedFromScratch` gets called. In an ideal world we would have a separate constructor for the `onLoad` case and it would be ok to call `init()` here but the way Jenkins works we have to call `init()` from the `onCreatedFromScratch`, `onCopiedFrom` and `onLoad` methods so that the field values are correctly populated first

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -0,0 +1,746 @@
...
+
+    @Override
+    public void onLoad(ItemGroup<? extends Item> parent, String name) throws IOException {
+        super.onLoad(parent, name);
+        init();
+        final Thread t = Thread.currentThread();
+        String n = t.getName();
+        try {
+            if (items ==null) {
+                // when Jenkins is getting reloaed, we want children being loaded to be able to find
```
stephenc: typo in `reloaed` also is the comment still valid? Does `loadChildren` respect the existing entries?

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
stephenc: because this is from `.../jobs/oldName/rename` and we want to redirect to `.../jobs/newName` not `.../jobs/oldName/newName`

stephenc:
:bee:

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
jglick: Yeah there is no good way to handle overridable methods to be called from Java constructors. My IDE at least will warn you if you seem to have gotten it wrong.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
jglick: Right, but the very existence of this field is provisional—I would like to delete it. If we are going to keep full settings compatibility for historical multibranch projects (discussed elsewhere), this field name would need to remain.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
jglick: Yeah it is the `save` which could fail, in which case the field is still initialized.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderCron.java
jglick: If it were thrown, you would want things to break visibly because something is wrong, which rethrowing the exception from an `Extension` constructor accomplishes.

jglick: FYI, this and most other code in these PRs already existed for years, it is just being moved around.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ThrottleComputationQueueTaskDispatcher.java
jglick: `isConcurrentBuild` returns `false` which should prevent concurrent indexing for a _single_ folder. This QTD prevents concurrent indexing of more than five _different_ folders.

jglick: As usual, this is simply a generalization of code which already existed for `MultiBranchProject`.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
jglick: I have no idea. It used to be on `Folder`, I was just pulling some code up.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
jglick: No, this is a helper method, intended to be overridden (so yes used, in other plugins), not a web method.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ChildObserver.java
jglick: `shouldUpdate` actually indicates a potential event if it returns non-null (if the subclass cares, which in fact it does not); both that and `mayCreate` are basically setup for `created`, which is definitely an event method. If Java had algebraic types like Scala I would have written something like this (which probably does not compile since I do not really know Scala):

```scala
abstract class Status
case class Existing(item: I) extends Status
case class Duplicate extends Status
case class Create(callback: Callback) extends Status
abstract class Callback {
  def created(child: I): unit
}
def observe(name: String): Status
```

and then a folder would write something like

```scala
observer.observe(projectName) match {
  case Existing(item) => item.updateSomehow()
  case Duplicate => // never mind
  case Create(callback) => callback.created(new …)
}
```

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
amuniz: But once the class name has changed, is possible to keep backward compatibility in settings?

jglick: Yes, using `addCompatibilityAlias`; see `PeriodicFolderTrigger` for an example. The difficulty with `OrphanedItemStrategy` is that the former field referring to the former class also had a different name (`deadBranchStrategy`), and there is no simple way to rename fields. You can do it with some work with a `readResolve` method. I intended to put this off until there is agreement from @gusreiber on whether we even want to retain the same config UI as was previously used, or switch to a different style that would require a slightly more complex migration. Also until I find the time to write some `@LocalData` tests proving that the implementation works as intended.

stephenc:
:bee:

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -0,0 +1,751 @@
...
+     * To add properties, use {@link #addProperty}.
+     */
+    public DescribableList<AbstractFolderProperty<?>,AbstractFolderPropertyDescriptor> getProperties() {
+        return properties;
+    }
+
+    @SuppressWarnings("rawtypes") // else setOwner will not compile
+    public void addProperty(AbstractFolderProperty p) throws IOException {
+        if (!p.getDescriptor().isApplicable(getClass())) {
+            throw new IllegalArgumentException(p.getClass().getName() + " cannot be applied to " + getClass().getName());
```
jglick: Note that this is stricter than the original version in `Folder`, which is a potential compatibility issue.

PR #21, opened 2015-10-14
[JENKINS-30951] implement ModifiableViewGroup instead of ViewGroup

daspilker:


jglick:
Will need a trivial merge with #20, also adding `@Override`.

daspilker:
Rebased to master after #20 got merged, added `@Override`.

jglick:
That was fast! :-)

PR #22, opened 2015-11-04
ComputedFolder.addTrigger

jglick:
@reviewbybees esp. @stephenc

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
@@ -248,6 +248,16 @@ public abstract class ComputedFolder<I extends TopLevelItem> extends AbstractFol
         return triggers.toMap();
     }
 
+    public void addTrigger(Trigger trigger) {
+        Trigger old = triggers.get(trigger.getDescriptor());
+        if (old != null) {
+            old.stop();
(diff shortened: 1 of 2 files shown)


stephenc:
:bee:

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
```diff
@@ -248,6 +248,16 @@ protected void submit(StaplerRequest req, StaplerResponse rsp) throws IOExceptio
...
     }
 
+    public void addTrigger(Trigger trigger) {
+        Trigger old = triggers.get(trigger.getDescriptor());
+        if (old != null) {
+            old.stop();
+            triggers.remove(old);
+        }
+        triggers.add(trigger);
+        trigger.start(this, true);
```
oleg-nenashev: Should it be explicitly saved here?
/me wonders about the <code>@PersistsData</code> annotation

jglick: > Should it be explicitly saved here?

`triggers` is a `DescribableList` and thus a `PersistedList`.

> /me wonders about [a hypothetical?] `@PersistsData` annotation

IMO annotations are of little value unless there is code which can somehow interpret them, at least during static analysis.

PR #23, opened 2015-11-04
[JENKINS-31129] Call scheduleBuild only after doConfigSubmit, not any save

jglick:
[JENKINS-31129](https://issues.jenkins-ci.org/browse/JENKINS-31129)

@reviewbybees esp. @stephenc


stephenc:
:bee:

PR #27, opened 2015-11-17
FolderProperty.owner binary compatibility fix

jglick:
Some (proprietary) plugins with `FolderProperty`s compiled against 4.x (prior to #20) were throwing `java.lang.NoSuchFieldError: owner`.

Testing:

    mvn -DskipTests clean install; for p in pre around post; do mvn -f src/test/folder-property-plugins/$p-5 -DskipTests clean package; cp -v src/test/folder-property-plugins/$p-5/target/$p-5.hpi src/test/resources/plugins; done; mvn -Dtest=FolderPropertyTest test

@reviewbybees esp. @stephenc


jglick:
Does not seem to work in a live environment; trying to figure out why.

jglick:
Hmm,

    mvn -Dtest=FolderPropertyTest test

passes but

    mvn -Dtest=FolderPropertyTest\#pre5 test

fails, meaning `@WithPlugin` does not work as advertised.

jglick:
[No glot…C’lom Fliday](https://github.com/jenkinsci/jenkins/blob/e8f2002c466c49eff4a7be0cc4633580515cd7a5/core/src/main/java/hudson/PluginManager.java#L347-L352)

jglick:
Possibly could still fix by calling `Jenkins.instance.pluginManager.compatibilityTransformer.loadRules(ClassLoader)` from an `@Initializer` in `FolderProperty` attaining `PLUGINS_LISTED`, though the loader must probably mask its parent so as to avoid double-parsing those in core. Getting increasingly scary.

Annoying that `JenkinsRule` has no option to load plugins in the test classpath (especially the “current” one) via `PluginManager`, which would allow the tests to actually reproduce the observed problem.

stephenc:
I fear `@AdaptField` on a field called `owner` is just critically too risky WDYT? @jtnord

jtnord:
I fear that is bound to bite someone in the arse.
There are fields called this in the following classes:
* hudson.model.Executor
* hudson.model.JobProperty
* hudson.model.View
* hudson.security.Permission
* hudson.security.PermissionGroup
* hudson.util.PersistedList
* hudson.widgets.HistoryWidget
let alone any plugins extending from this and adding their own ownser fields (RunAction2 etc)?

So until [JENKINS-28799 ](https://issues.jenkins-ci.org/browse/JENKINS-28799) is fixed I would run a mile from something as scary as this.

jglick:
I did not realize JENKINS-28799 was still outstanding. In that case I agree, this is way too risky.

PR #28, opened 2015-12-11
FIXED JENKINS-28043 reload jobs config

fengxx:
[JENKINS-28043](https://issues.jenkins-ci.org/browse/JENKINS-28043)

reload is not working after commit https://github.com/jenkinsci/jenkins/commit/9fc87f4a22ef4e32445451f5114fb800b6b73e9e in Jenkins core because 

```
V item = (V) parent.getItem(subdir.getName()); 
```

will always get the old item, the items was set to old set by 

```
this.items = ((AbstractFolder) current).items; 
```

in cloudbees-folder-plugin

The patch fixed the issue

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
@@ -70,6 +70,7 @@ import hudson.util.HttpResponses;
 import hudson.views.DefaultViewsTabBar;
 import hudson.views.ViewsTabBar;
 import java.io.File;
+import java.io.FileFilter;
 import java.io.IOException;
 import java.net.URLEncoder;
 import java.text.ParseException;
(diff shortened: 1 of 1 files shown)


p-zilla:
@fengxx: in your comment you said reload is work, does that mean this patch requires Jenkins core that incorporates commit 9fc87f4a22ef4e32445451f5114fb800b6b73e9e? Or does this patch not require that?

fengxx:
update comment to mke it more clear

jglick:
There is no test here.

The patch is not right, though it may be fixing this bug. Probably requires more investigation, and perhaps the fix needs to be in core.

jglick:
Judging by [this original code](https://github.com/jenkinsci/cloudbees-folder-plugin/blob/869ef0ce662d643aa16f3ce55150a34f0c7b41d2/src/main/java/com/cloudbees/hudson/plugins/folder/Folder.java#L230-L254) this has long been broken—`items` is simply set twice.

oleg-nenashev:
@fengxx Any plans regarding this PR? Seems it requires tests at least

fengxx:
I will check JenkinsRule to write a testcase for coverage

jglick:
Seems to be abandoned. Judging by recent-ish comments in Jira, the described bug may still exist.

PR #29, opened 2015-12-26
Add a new "Ball Color Icon" as an additional Icon option.

mjdetullio:
Primarily to be consumed by multibranch projects.

Added the getBallColor method to AbstractFolder so subclasses can override the method if desired.

Alternative implementation to #17

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
@@ -38,6 +38,7 @@ import hudson.init.Initializer;
 import hudson.model.AbstractItem;
 import hudson.model.Action;
 import hudson.model.AllView;
+import hudson.model.BallColor;
 import hudson.model.Descriptor;
 import hudson.model.HealthReport;
 import hudson.model.Item;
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
LGTM :+1:

jbkempf:
Is there any news on this merge?

oleg-nenashev:
@jbkempf As far as I know @jglick wants to perform a bulk review of icon-related PRs, because he has quite different implementation in his drafts. I'll assign this to him.

mjdetullio:
Ultimately I put a different implementation in the Multi-Branch Project Plugin which does not require any changes to the folder type.  It's closer to PR #17 as I mentioned when I opened this.  https://github.com/jenkinsci/multi-branch-project-plugin/blob/multi-branch-project-plugin-0.4.2/src/main/java/com/github/mjdetullio/jenkins/plugins/multibranch/BallColorFolderIcon.java

I was under the impression there were no plans to merge any "Ball Color" icons since the implementation is in the "Folders Plus" plugin.  Merging any of them would be a conflict of CloudBees' interests.

With the current API it's possible to create additional plugins to add colors without having to get them merged here.

For these reasons I'll be closing this one, however if Jessie wants this for any reason then it can be reopened.

jbkempf:
Well, if you have different items that are not different branches of the same project, that does not work, does it?

mjdetullio:
#getAllJobs will recursively retrieve all items that inherit Job, regardless of what those items represent.  So actually it does work outside of multi-branch types, including folders with nested folders.

PR #32, opened 2016-01-08
[FIXED JENKINS-32359] - Properly assign persisted list owners when loading folders from disk

oleg-nenashev:
The issue causes a critical bug in CloudBees Folders API. I suppose it can be reproduced manually with plugin-internal properties if you follow the unit test.
https://issues.jenkins-ci.org/browse/JENKINS-32359

@reviewbybees


rsandell:
:bee:

Vlatombe:
:bee:

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -184,6 +184,8 @@ protected AbstractFolder(ItemGroup parent, String name) {
     protected void init() {
         if (properties == null) {
             properties = new DescribableList<AbstractFolderProperty<?>,AbstractFolderPropertyDescriptor>(this);
+        } else {
+            properties.setOwner(this);
```
jtnord: why is this done on `init` (called from `onLoad()`) rather than in readResolve?

oleg-nenashev: Maybe it's a lazy hack allowing to call the same method from constructors

src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
```diff
@@ -278,6 +286,32 @@ private void copyFromGUI(Folder f, JenkinsRule.WebClient wc, String fromName, St
...
     }
+    
+    @Issue("JENKINS-32359")
+    @Test public void shouldProperlyPersistFolderPropertiesOnMultipleReloads() throws Exception {
+        Folder folder = r.jenkins.createProject(Folder.class, "myFolder");
+        r.jenkins.setAuthorizationStrategy(new ProjectMatrixAuthorizationStrategy());
+        
+        // We add a stub property to generate the persisted list
+        // After that we save and reload the config in order to drop PersistedListOwner according to the JENKINS-32359 scenario
+        folder.addProperty(new FolderCredentialsProvider.FolderCredentialsProperty(new DomainCredentials[0]));
```
jtnord: what is this first part doing that is important to the test that the second part is not doing?

i.e. why 2 properties and 2 reloads - yet you are only asserting on one property being persisted?

oleg-nenashev: I need to save a non-empty properties list to the disk and reload it. It makes the code to bypass the <code>if (properties == null)</code> clause

oleg-nenashev: At the end of the day, that's how I've got into it during https://github.com/jenkinsci/ownership-plugin/pull/44 testing :)

jtnord:
:bee:

oleg-nenashev:
@reviewbybees done

stephenc:
:bee: :+1:

oleg-nenashev:
Another TC is available here: https://github.com/oleg-nenashev/ownership-plugin/commit/b42f2937c1f91bfd521e00650cbee38cd565c6ba

jglick:
Analogous to #19 IIUC.

oleg-nenashev:
@jglick 
The title of this PR seemed to be unrelevant, so I didn't even open it :(
BTW this PR also has Unit tests

jglick:
@oleg-nenashev well #19 purports to fix ownership of `FolderIcon`, whereas you are trying to fix ownership of `FolderProperty`.

oleg-nenashev:
@jglick Gentle bump, any chance to get a release on this week? It blocks https://github.com/jenkinsci/ownership-plugin/pull/44

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
jglick: During `readResolve`, `this` will in general be incomplete in various ways. Unsafe to pass an instance to plugin code which might expect to do things like call `owner.getParent()`, etc. `onLoad` is called later when everything is wired up.

src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
```diff
@@ -278,6 +286,32 @@ private void copyFromGUI(Folder f, JenkinsRule.WebClient wc, String fromName, St
...
+        folder.doReload();
```
jglick: :bug: This is not convincing. You are retaining the local variable reference to the original `Folder` object, unlike the real situation where a new instance is constructed.

Rather use `Jenkins.reload` and call `getItemByFullName` to find the folder again; or use `RestartableJenkinsRule` in a new test suite.

Fixing this may also allow you to get rid of the distracting reference to `FolderCredentialsProperty`, which as @jtnord pointed out has no clear purpose in this test.

jglick:
> you are trying to fix ownership of `FolderProperty`

I was misled. `FolderProperty.owner` already was set. You are rather fixing the ownership of the property _list_, relevant only for resaving the folder `config.xml` when making changes to that list.

src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
```diff
@@ -278,6 +286,32 @@ private void copyFromGUI(Folder f, JenkinsRule.WebClient wc, String fromName, St
...
+        grantedPermissions.put(Jenkins.ADMINISTER, sids);
+        folder.addProperty(new com.cloudbees.hudson.plugins.folder.properties.AuthorizationMatrixProperty(grantedPermissions));
+        
+        // Reload folder from disk and check the state
+        folder.doReload();
+        Folder reloadedFolder = r.jenkins.getItemByFullName("myFolder", Folder.class);
+        assertThat("Folder has not been found after the reloading", reloadedFolder, notNullValue());
+        assertThat("Property has not been reloaded, hence it has not been saved properly",
+            reloadedFolder.getProperties().get(com.cloudbees.hudson.plugins.folder.properties.AuthorizationMatrixProperty.class),
+            notNullValue());
```
jglick: Would be nice to also verify that the `AbstractFolderProperty.owner` is set on all properties after reload.

Since that is `protected`, and not trivially accessed from implementations in this plugin, may be easier to use `@TestExtension` properties for this purpose.

src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
oleg-nenashev: ok

src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
oleg-nenashev: I see nothing bad in adding a public getter to AbstractFolder Property.
BTW it can be reduced to package-internal Restricted method

oleg-nenashev:
@jglick done

oleg-nenashev:
@jglick bump

src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
```diff
@@ -278,7 +286,71 @@ private void copyFromGUI(Folder f, JenkinsRule.WebClient wc, String fromName, St
             }
         });
     }
+    
+    @Issue("JENKINS-32487")
+    @Test public void shouldAssignPropertyOwnerOnCreationAndReload() throws Exception {
+        Folder folder = r.jenkins.createProject(Folder.class, "myFolder");
+        r.jenkins.setAuthorizationStrategy(new ProjectMatrixAuthorizationStrategy());
```
jglick: Clashes with #33.

jglick: Use a mock instead?

jglick:
Unmergeable @oleg-nenashev.

oleg-nenashev:
@jglick Yes, I should spend some time and finally integrate it

pom.xml
```diff
@@ -56,6 +56,13 @@
       <version>2.0.4</version>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>matrix-auth</artifactId>
+      <version>1.3</version>
+      <scope>test</scope>
+      <type>jar</type>
```
jglick: redundant

pom.xml
```diff
@@ -56,6 +56,13 @@
       <version>2.0.4</version>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>matrix-auth</artifactId>
+      <version>1.3</version>
+      <scope>test</scope>
```
jglick: Not exactly happy to see this come back, but I guess in `test` scope it is OK, and you needed two distinct `AbstractFolderProperty` implementations I suppose?

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderProperty.java
```diff
@@ -54,10 +54,21 @@
...
-    protected void setOwner(C owner) {
+    protected void setOwner(@NonNull C owner) {
         this.owner = owner;
     }
 
+    /**
+     * Gets an owner of the property.
+     * @return Owner of the property.
+     *         It should be always non-null if the property initialized correctly.
+     * @since TODO
```
jglick: Would rather not see this be added to the API. There is no use case for it/.

jglick:
:bee: with some comments

pom.xml
oleg-nenashev: Yes. I could create Mock ones, but a realistic case slightly improves the test coverage

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderProperty.java
oleg-nenashev: I'll add restricted

jglick:
re:bee:

oleg-nenashev:
@reviewbybees done

PR #33, opened 2016-02-08
Removing matrix-auth dependency

jglick:
Finally solves the improper dependency, a holdover from when this plugin was not OSS.

@reviewbybees esp. @stephenc


pom.xml
```diff
@@ -17,6 +17,10 @@
     by project type, organization type etc). Folders are nestable and you can define views within folders.
   </description>
   <url>https://wiki.jenkins-ci.org/display/JENKINS/CloudBees+Folders+Plugin</url>
+  <properties>
+    <jenkins.version>1.609.1</jenkins.version>
+    <jenkins-test-harness.version>2.3-SNAPSHOT</jenkins-test-harness.version>
```
jglick: https://github.com/jenkinsci/jenkins-test-harness/pull/11

pom.xml
```diff
@@ -46,38 +50,16 @@
 
   <dependencies>
     <dependency>
-      <groupId>com.github.stephenc.findbugs</groupId>
-      <artifactId>findbugs-annotations</artifactId>
-      <version>1.3.9-1</version>
-      <scope>provided</scope>
-      <optional>true</optional>
-    </dependency>
-    <dependency>
```
jglick: Cleanup from new parent POM.

src/main/java/com/cloudbees/hudson/plugins/folder/Folder.java
```diff
@@ -94,7 +94,7 @@
     /**
      * {@link Action}s contributed from subsidiary objects associated with
      * {@link Folder}, such as from properties.
-     * <p/>
+     * <p>
```
jglick: JDK 8 `javadoc` fixes, forced by new parent POM.

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
```diff
@@ -21,6 +21,7 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 -->
+<?jelly escape-by-default='true'?>
```
jglick: Forced by new `maven-hpi-plugin` from new parent POM.

src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
```diff
@@ -246,17 +245,13 @@ private void copyFromGUI(Folder f, JenkinsRule.WebClient wc, String fromName, St
...
     @Test public void discoverPermission() throws Exception {
         r.jenkins.setSecurityRealm(r.createDummySecurityRealm());
-        ProjectMatrixAuthorizationStrategy as = new ProjectMatrixAuthorizationStrategy();
-        as.add(Jenkins.READ, "anonymous");
-        as.add(Jenkins.READ, "authenticated");
-        as.add(Item.DISCOVER, "authenticated");
-        r.jenkins.setAuthorizationStrategy(as);
         final Folder d = r.jenkins.createProject(Folder.class, "d");
-        /* Cannot make test be meaningful unless using matrix-auth 1.2 and setting blocksInheritance on h.s.AMP on both p1 & p2:
-        d.addProperty(new com.cloudbees.hudson.plugins.folder.properties.AuthorizationMatrixProperty(Collections.singletonMap(Item.READ, new HashSet<String>(Arrays.asList("anonymous", "authenticated")))));
```
jglick: Actually done correctly now, by using `onItems(d)` rather than `onFolders(d)`

oleg-nenashev:
:bug: since there is not explicit requirement to update matrix-auth plugin. At least it may break JobDSL plugin (CC @daspilker ).

My proposals:
* Add incompatibleSince warning to pom.xml
* Add update guidelines to the plugin's Wiki page

daspilker:
This will not necessarily break Job DSL, but existing and newly generated configs if user's do not upgrade matrix-auth. So setting incompatibleSince and putting a sign on the wiki seems to be a good idea.

jglick:
> it may break JobDSL plugin

It should not: the class name is unmodified.

That said, `incompatibleSince` is useful as a warning to users to update both plugins.

oleg-nenashev:
:bee:

svanoort:
:bee: for me

PR #38, opened 2016-03-10
Display names has been changed

recena:
@reviewbybees specially @rtyler and @daniel-beck

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -12,9 +12,9 @@
   <version>5.4-SNAPSHOT</version>
   <packaging>hpi</packaging>
 
-  <name>CloudBees Folders Plugin</name>
+  <name>Folders Plugin</name>
   <description>This plugin allows users to create "folders" to organize jobs. Users can define custom taxonomies (like
-    by project type, organization type etc). Folders are nestable and you can define views within folders.
+    by project type, organization type etc). Folders are nestable and you can define views within folders. Maintained by CloudBees, Inc.
   </description>
   <url>https://wiki.jenkins-ci.org/display/JENKINS/CloudBees+Folders+Plugin</url>
   <properties>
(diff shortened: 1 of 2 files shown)


kwhetstone:
:+1: I like this (I'll let Tyler and Daniel weigh in for the bees)

HRMPW:
:bee:

HRMPW:
"Maintained" or "Contributed" both sound fine for me but let's be consistent. If we're going to use "Maintained by CloudBees" we should use the same phrase on all of them.

jglick:
:bee:, long overdue

stephenc:
:bee: w00t!

oleg-nenashev:
:bee: . I won't need to explain it during demos anymore :+1:

recena:
@reviewbybees done

pom.xml
```diff
@@ -12,9 +12,9 @@
   <version>5.4-SNAPSHOT</version>
   <packaging>hpi</packaging>
 
-  <name>CloudBees Folders Plugin</name>
+  <name>Folders Plugin</name>
   <description>This plugin allows users to create "folders" to organize jobs. Users can define custom taxonomies (like
-    by project type, organization type etc). Folders are nestable and you can define views within folders.
+    by project type, organization type etc). Folders are nestable and you can define views within folders. Maintained by CloudBees, Inc.
   </description>
   <url>https://wiki.jenkins-ci.org/display/JENKINS/CloudBees+Folders+Plugin</url>
```
jglick: @recena do you not plan to also rename the wiki pages?

recena: @jglick Yes, but after to Jenkins 2.0. Is it ok for you?

jglick: You mean, because you are too busy now? That is fine, just checking if it was planned.

recena: @jglick Exactly. I prefer to do all pending things at the same time: cloudbees-disk-usage-simple-plugin, cloudbees-folder-plugin, etc...

PR #40, opened 2016-03-14
Allow Delete Folder to appear in context menu

jglick:
So long as [JENKINS-32585](https://issues.jenkins-ci.org/browse/JENKINS-32585) is not implemented, a link in the sidepanel ought to be in the context menu unless we have explicitly specified `contextMenu=false`, which in this case we have not. (Anyway I find it quite convenient to use the context menu to blow away unwanted folders!)

The problem was that [here](https://github.com/jenkinsci/cloudbees-folder-plugin/blob/cd19df8330048af010056ce9c5d0920516632c4b/src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/tasks-new.jelly#L29) `view` was undefined.

@reviewbybees

### DIFF
--- a/src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/sidepanel.jelly
+++ b/src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/sidepanel.jelly
@@ -22,4 +22,9 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 -->
 <?jelly escape-by-default='true'?>
-<st:include it="${it.primaryView}" page="sidepanel.jelly" xmlns:st="jelly:stapler"/>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler">
+    <j:if test="${view == null}"> <!-- true when rendering from ModelObjectWithContextMenu.ContextMenu.from -->
+        <j:set var="view" value="${it.primaryView}"/>
+    </j:if>
+    <st:include it="${it.primaryView}" page="sidepanel.jelly"/>
+</j:jelly>


src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/sidepanel.jelly
```diff
@@ -22,4 +22,9 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 -->
 <?jelly escape-by-default='true'?>
-<st:include it="${it.primaryView}" page="sidepanel.jelly" xmlns:st="jelly:stapler"/>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler">
+    <j:if test="${view == null}"> <!-- true when rendering from ModelObjectWithContextMenu.ContextMenu.from -->
```
kzantow: Is this a bug in MOWCM ?

kzantow:
:bee:

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/sidepanel.jelly
```diff
@@ -22,4 +22,9 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
+        <j:set var="view" value="${it.primaryView}"/>
+    </j:if>
+    <st:include it="${it.primaryView}" page="sidepanel.jelly"/>
```
amuniz: Is this including itself?! I mean, this file is `sidepanel.jelly` and this line is including `sidepanel.jelly`...

amuniz:
:bee: as the include recursion should work because it was there before (I must be missing something).

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/sidepanel.jelly
jglick: Not really, it is just going directly to the object’s `sidepanel` view, whereas a regular folder index page load actually loads `View/sidepanel.jelly` via [this trick](https://github.com/jenkinsci/cloudbees-folder-plugin/blob/5f239712601807f3f1eb6f573a8fa550261978a7/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java#L416-L422).

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/sidepanel.jelly
jglick: `AbstractFolder/sidepanel.jelly` is loading `View/sidepanel.jelly`.

PR #47, opened 2016-03-25
[JENKINS-33815] Fixing FolderComputation to properly handle AbortException

jglick:
Needed to do https://github.com/jenkinsci/github-branch-source-plugin/pull/35 properly.

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderComputation.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderComputation.java
@@ -24,11 +24,11 @@
 
 package com.cloudbees.hudson.plugins.folder.computed;
 
+import hudson.AbortException;
 import hudson.BulkChange;
 import hudson.Util;
 import hudson.XmlFile;
 import hudson.console.AnnotatedLargeText;
-import hudson.model.AbstractItem;
 import hudson.model.Actionable;
 import hudson.model.BallColor;
 import hudson.model.Cause;
@@ -41,7 +41,6 @@ import hudson.model.Saveable;
 import hudson.model.StreamBuildListener;
 import hudson.model.TopLevelItem;
 import hudson.model.listeners.SaveableListener;
-import hudson.model.queue.SubTask;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
:bee:

src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderComputation.java
```diff
@@ -113,8 +112,12 @@ public void run() {
             listener.getLogger().println("Aborted");
             _result = Result.ABORTED;
         } catch (Exception x) {
-            // TODO skip stack trace for AbortException
-            x.printStackTrace(listener.fatalError("Failed to recompute children of " + getDisplayName()));
+            String message = "Failed to recompute children of " + folder.getFullDisplayName();
+            if (x instanceof AbortException) {
+                listener.fatalError(message + ": " + x.getMessage());
```
recena: @jglick In this way your are **forcing** the message with *Failed to recompute children of*. I prefer something like that:

```
listener.fatalError("%n" + x.getMessage() + "%n%n");
```

recena: @jglick Or, I would use a `LOGGER`...

```
String message = "Failed to recompute children of " + folder.getFullDisplayName();
LOGGER.log(Level.WARNING, message + ": " + x.getMessage(), x)
listener.fatalError(x.getMessage());
```

src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderComputation.java
```diff
@@ -110,11 +109,16 @@ public void run() {
...
             listener.getLogger().println("Aborted");
             _result = Result.ABORTED;
         } catch (Exception x) {
-            // TODO skip stack trace for AbortException
-            x.printStackTrace(listener.fatalError("Failed to recompute children of " + getDisplayName()));
+            LOGGER.log(Level.FINE, "recomputation of " + folder.getFullName() + " failed", x);
+            if (x instanceof AbortException) {
+                listener.fatalError(x.getMessage());
+            } else {
+                x.printStackTrace(listener.fatalError("Failed to recompute children of " + folder.getFullDisplayName()));
```
recena: @jglick Why are you using `folder.getFullDisplayName()` and `folder.getFullName()`? Both return the same?

recena:
:bee:

src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderComputation.java
jglick: No, `fullName` would be for example `myorg/feature%2Fstuff` where `fullDisplayName` might be `My Org » feature/stuff`.

recena: Thanks.

PR #48, opened 2016-03-25
[JENKINS-33759] Use standard `newJob` view, not cloned `new` view

jglick:
[JENKINS-33759](https://issues.jenkins-ci.org/browse/JENKINS-33759)

At root level, the [sidepanel was already linking](https://github.com/jenkinsci/jenkins/blob/320ff5ff446f84657ba88c445260e1a8ea43ffdb/core/src/main/resources/hudson/model/View/sidepanel.jelly#L43) to `View/newJob`. And `View/noJob` and its various overrides was too. But `Folder` in its sidepanel was overriding `tasks-new` and, via `tasks-create`, was linking to its own `new` view. In Jenkins 1.x that looks the same as `newJob`, but in 2.x `newJob` was modified, so the difference is apparent: you would get the new UI when clicking in the main area of an empty view in a newly created folder, but the old UI if you clicked _New Item_ from the sidepanel.

@reviewbybees


jglick:
Still working on some additional testing and cleanup in core.

recena:
:bee:

recena:
@jglick Which cleanup in core?

jglick:
Working on it, you will see…

src/main/java/com/cloudbees/hudson/plugins/folder/Folder.java
```diff
@@ -373,59 +370,6 @@ public TopLevelItem newInstance(ItemGroup parent, String name) {
             return new Folder(parent, name);
         }
 
-        /**
-         * Auto-completion for the "copy from" field in the new job page.
-         */
-        public AutoCompletionCandidates doAutoCompleteCopyNewItemFrom(@AncestorInPath final Folder f,
```
jglick: Was not actually being used, it turns out, and core is a better place to fix the details.

src/main/java/com/cloudbees/hudson/plugins/folder/Folder.java
```diff
@@ -373,59 +370,6 @@ public TopLevelItem newInstance(ItemGroup parent, String name) {
...
-                    super.onItemGroup(group);
-                }
-
-                @Override
-                public void onItem(Item i) {
-                    String name = getPathNameOf(i);
-                    if (name.startsWith(value)) {
-                        if (i instanceof TopLevelItem) {
-                            TopLevelItem tli = (TopLevelItem) i;
-                            if (f.isAllowedChild(tli)) {
```
jglick: Now working in core via https://github.com/jenkinsci/jenkins/pull/2173.

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/tasks-new.properties
```diff
@@ -1,2 +1 @@
 delete=Delete {0}
-newJob=New {0}
```
jglick: Seems to have been unused.

andresrc:
:bee:

christ66:
:bee:

oleg-nenashev:
:bee:

PR #49, opened 2016-03-25
Meaningful FolderComputation.toString

jglick:
Otherwise when diagnosing issues like that perhaps fixed in https://github.com/jenkinsci/subversion-plugin/pull/158, you see thread dumps with thread names like

    Executor #-1 for master : executing jenkins.branch.MultiBranchProject$BranchIndexing@1234abcd

which is not very helpful.

@reviewbybees


oleg-nenashev:
:bee:

andresrc:
:bee:

PR #50, opened 2016-03-29
[JENKINS-33817] Option to save backup copies of computation log

jglick:
Minimal solution of [JENKINS-33817](https://issues.jenkins-ci.org/browse/JENKINS-33817), sufficient for diagnosing recomputation bugs. For now, no UI option to enable, UI browsing of old logs, or smart retention policy (like saving only logs that orphaned children).

@reviewbybees


amuniz:
I think a UI configuration option with a sensible default value would be much better, but probably out of scope, so :bee:

recena:
I agree with @amuniz. *this kind of solutions are for whole life*

:bee:

PR #52, opened 2016-03-31
Expect FolderAddFilter to be called from View.doCategories

jglick:
Fixes a bug identified in https://github.com/jenkinsci/jenkins/pull/2191.

Merely skipping the URI check, and applying the filtering unconditionally given a folder context, would seem simpler, but this would break code which is not part of _New Item_. For example, `SubItemFilterProperty` in `cloudbees-folders-plus` has a configuration fragment which lets you select item types to filter out. If `FolderAddFilter` were applied in `/job/folder/configure`, once you limited the folder to one item type, you would not be able to undo your decision, because the others would no longer be shown!

Tested using `SubItemFilterProperty` against both Jenkins 1.x and 2.x.

@reviewbybees esp. @recena

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/FolderAddFilter.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/FolderAddFilter.java
@@ -40,7 +40,8 @@ import org.kohsuke.stapler.StaplerRequest;
 
     @Override public boolean filter(Object context, Descriptor descriptor) {
         StaplerRequest req = Stapler.getCurrentRequest();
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
:bee: excepting the fact I'm still against /categories

jglick:
> I'm still against /categories

Not sure what you mean by that but I suppose it is a topic for another PR?

oleg-nenashev:
> Not sure what you mean by that but I suppose it is a topic for another PR?

right

recena:
:bee: 

It seems difficult to manage all of these URL paths.

recena:
There are two URL paths today. But if we create more services....?

PR #46, opened 2016-03-25
[JENKINS-25240] Need to consider even nested jobs for purposes of checking running builds

jglick:
[JENKINS-25240](https://issues.jenkins-ci.org/browse/JENKINS-25240), again.

@reviewbybees

PR #44, opened 2016-03-16
[JENKINS-25240] Do not delete branch projects with running builds

jglick:
[JENKINS-25240](https://issues.jenkins-ci.org/browse/JENKINS-25240)

@reviewbybees

PR #53, opened 2016-03-31
[JENKINS-25240] IllegalStateException from a list iterator

jglick:
A bug that was technically introduced in #44 but made much more likely in #46.

@reviewbybees


oleg-nenashev:
ohhh, :bee:

PR #56, opened 2016-04-04
[JENKINS-31162] Support for Item categorization in Jenkins 2

recena:
This PR belongs to this [JENKINS-31162](https://issues.jenkins-ci.org/browse/JENKINS-31162) issue and allows to categorize the Folders in a right way.

downstream of https://github.com/jenkinsci/jenkins/pull/2217

@reviewbybees


src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
```diff
@@ -41,6 +41,40 @@ public String getDisplayName() {
...
+
+    /**
+     * Needed if it wants Folder are categorized in Jenkins 2.x.
+     *
+     * TODO: Override when the baseline is upgraded to 2.x
+     *
+     * @return A string it represents a URL pattern to get the Item icon in different sizes.
+     */
+    public String getIconFilePathPattern() {
+        return "plugin/cloudbees-folder/images/:size/folder.png";
```
jglick: Why are you not just using the `images/:size/folder.png` in Jenkins core?

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
```diff
@@ -41,6 +41,40 @@ public String getDisplayName() {
     }
 
     /**
+     * Needed if it wants Folders are categorized in Jenkins 2.x.
+     *
+     * TODO: Override when the baseline is upgraded to 2.x
+     * TODO: Replace to {@code NestedProjectsCategory.ID}
```
jglick: Well I think once we have a 2.x dependency, we can just move that category definition here.

recena: AFAIK, It is not decided.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
recena: I prefer to have a dedicated resource since this plugin has his own requirements. Additionally, the graphical resources are different.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
recena: AFAIK, `NestedProjectsCategory` will be provide by core though it can be provided by this plugin and additionally we can test the new API. Are you ok?

jglick: > NestedProjectsCategory will be provide by core though it can be provided by this plugin

It is fine to _initially_ keep the category in core, so long as it is `@Restricted` so we retain the option of moving it into `cloudbees-folder` later.

recena: @jglick Why not now?

jglick: You mean, why not add the category to the plugin now? You could, if you wanted to make this plugin now depend on `2.0-beta-2` or similar, and switched its own version to a beta number. We would also need to make sure that the update center is set up for plugins with a 2.x core dep. In that case there is no need to put the category into core. But I see no particular advantage, since the main reason to do this would be to export a plugin API for the category ID, which means making _other_ plugins depend on the beta release of `cloudbees-folder` and thus core, so they in turn would need new beta releases. Probably too much work for now.

recena: Totally true. Thanks.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
```diff
@@ -41,6 +41,40 @@ public String getDisplayName() {
...
+
+    /**
+     * Needed if it wants Folders are categorized in Jenkins 2.x.
+     *
+     * TODO: Override when the baseline is upgraded to 2.x
+     *
+     * @return A string with the Item description.
+     */
+    public String getDescription() {
+        return Messages.Folder_Description();
```
jglick: :bug: this should not be implemented here, but rather left abstract and implemented in `Folder.DescriptorImpl`.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
```diff
@@ -41,6 +41,40 @@ public String getDisplayName() {
...
+    }
+
+    /**
+     * Needed if it wants Folder are categorized in Jenkins 2.x.
+     *
+     * TODO: Override when the baseline is upgraded to 2.x
+     *
+     * @return A string it represents a URL pattern to get the Item icon in different sizes.
+     */
+    public String getIconFilePathPattern() {
```
jglick: :bug: this also should probably be left abstract and implemented in `Folder.DescriptorImpl`, unless you intend to set a default to be used for all `AbstractFolder` subtypes unless overridden.

recena: This was my intention. For that reason, I don't understand your bugs.

recena: @jglick I addressed your comments [here](https://github.com/jenkinsci/cloudbees-folder-plugin/pull/56/commits/9f8fa69e7386a33510ad900e577094ef455d08d6)

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
amuniz: > Well I think once we have a 2.x dependency, we can just move that category definition here.

Yes, and the javadoc will still be valid I think (if we keep the class name).

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
amuniz: If it becomes abstract, would not be backward incompatible?

amuniz:
I left some comments in the files view, but I can not see them now... GH is not behaving well today.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
jglick: True, I should rather have said that it should be left undefined.

jglick:
> I left some comments in the files view, but I can not see them now

The PR UI was overhauled a couple weeks ago. If the comments still apply to the current diff they should be displayed. But I see some comments of yours in an outdated diff.

At any rate, :bee:

recena:
@jglick your re-bee would be great.

@amuniz your bug or bee would be great.

jglick:
re:bee:

amuniz:
:bee:

recena:
@reviewbybees done

PR #57, opened 2016-04-18
[FIXED JENKINS-34306] When moving an item several destinations showed in the GUI are not valid

fbelzunc:
1. create a job A in a folder A 
2. create a job with the same name A in another folder B 
3. Then, the move option tells you can move to:

```
Jenkins
Jenkins >> Folder A
Jenkins >> Folder B
````

But this is wrong because `Jenkins >> Folder A` is the item itself and in `Jenkins >> Folder B` there is already an item with the same name - we should cover these cases.

https://issues.jenkins-ci.org/browse/JENKINS-34306

@reviewbybees CC @jglick

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/relocate/StandardHandler.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/relocate/StandardHandler.java
@@ -86,31 +86,34 @@ import org.kohsuke.stapler.HttpResponses;
                     continue;
                 }
                 ItemGroup<?> p = itemGroup;
-                Item i = (Item) p;
-                if (i == item || i == item.getParent()) {
-                    // Cannot move a folder into itself or a descendant.
-                    // Cannot move an item into the same path where it is
-                    continue ITEM;
-                }
-                if (i instanceof Folder) {
-                    Folder folder = (Folder) i;
-                    if (folder.getItem(item.getName()) != null) {
-                        // Cannot move an item into a Folder if there is already an item with the same name
+
+                if (p instanceof Item) {
+                    Item i = (Item) p;
+                    if (i == item || i == item.getParent()) {
+                        // Cannot move a folder into itself or a descendant.
+                        // Cannot move an item into the same path where it is
                         continue ITEM;
                     }
-                }
-                // Cannot move d1/ into say d1/d2/d3/
-                ItemGroup itemGroupSubElement = i.getParent();
-                while (itemGroupSubElement != instance) {
-                    if (itemGroupSubElement instanceof Folder) {
-                        Folder currentFolder = (Folder)itemGroupSubElement;
-                        if (item == currentFolder) {
+                    if (i instanceof Folder) {
+                        Folder folder = (Folder) i;
+                        if (folder.getItem(item.getName()) != null) {
+                            // Cannot move an item into a Folder if there is already an item with the same name
                             continue ITEM;
                         }
-                        itemGroupSubElement = currentFolder.getParent();
                     }
+                    // Cannot move d1/ into say d1/d2/d3/
+                    ItemGroup itemGroupSubElement = i.getParent();
+                    while (itemGroupSubElement != instance) {
+                        if (itemGroupSubElement instanceof Folder) {
+                            Folder currentFolder = (Folder) itemGroupSubElement;
+                            if (item == currentFolder) {
+                                continue ITEM;
+                            }
+                            itemGroupSubElement = currentFolder.getParent();
+                        }
+                    }
+                    result.add(itemGroup);
                 }
-                result.add(itemGroup);
             }
         }
         return result;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by christ66** [comment]
:bug: You can't move into subfolders anymore:

If you have a project in B and create a subfolder:


The subfolder does not become an available option anymore:

**Comment by fbelzunc** [comment]
You are right!

I am fixing it.

**Comment by fbelzunc** [comment]
@christ66 Would you mind to review again, please?

**Comment by christ66** [comment]
Tested locally and I was able to move into sub-folders again. :bee:

File: src/main/java/com/cloudbees/hudson/plugins/folder/relocate/StandardHandler.java [hunk 1]
```diff
@@ -85,13 +86,18 @@ public HandlingMode applicability(Item item) {
                     continue;
                 }
                 ItemGroup<?> p = itemGroup;
-                while (p instanceof Item) {
-                    Item i = (Item) p;
-                    if (i == item) {
-                        // Cannot move a folder into itself or a descendant.
+                Item i = (Item) p;
```
lordofthejars: is normal to not having instanceof `Item` in this case?

jtnord:
I think there is a :bug: where it will [always say you can move the item to the root of the Jenkins](https://github.com/jenkinsci/cloudbees-folder-plugin/pull/57/files#diff-800a78b099f2e98dd244ebe63eebd391R80) instance even if there is a job with he same name already existing.

fbelzunc:
@jtnord You are right. A new test I created demonstrate you are right.

jglick:
@stephenc could say why, but it was intentional that the _current_ position was shown in the pulldown, as the initial selection.

Whether we should quietly hide a folder with an item of the same name, I am not sure. Probably better handled by letting you select that folder but then showing an explicit error.

The original code also blocked you from moving `d1/` into, say, `d1/d2/d3/`, which is a defense you seem to have lost.

fbelzunc:
> @stephenc could say why, but it was intentional that the current position was shown in the pulldown, as the initial selection.

I pinged him to understand why it was intended.

> Whether we should quietly hide a folder with an item of the same name, I am not sure. Probably better handled by letting you select that folder but then showing an explicit error.

From User's point of view my opinion is not to expose current position in the pulldown as it will not work - at least for this feature.

> Probably better handled by letting you select that folder but then showing an explicit error.

If something is an error, it shouldn't be exposed unless there is a good reason - waiting @stephenc to answer this. Mainly, because JENKINS-19584 will not block to actually perform the move operation.

> The original code also blocked you from moving d1/ into, say, d1/d2/d3/, which is a defense you seem to have lost.

This was corrected - I added a bunch of tests to add more robustness to this plugin.

stephenc:
> I pinged him to understand why it was intended.
 so that by default the move is a no-op if you hit it by mistake

fbelzunc:
I always need to read @stephenc 's comments 10 times to understand what he wants to tell me ;-) - so If I understand correctly there is not any good reason to keep the old default behaviour.

stephenc:
I think it is a good reason that the default does not move, just like if you open the config page and hit save, by default that will be a no-op unless you change something on the page

fbelzunc:
> I think it is a good reason that the default does not move, just like if you open the config page and hit save, by default that will be a no-op unless you change something on the page

Now I understood you ;-)

@jglick @stephenc @christ66 @jtnord  I think I have addressed all the comments now.

jtnord:
:bee:

stephenc:
:bee:

christ66:
:bee:

src/main/java/com/cloudbees/hudson/plugins/folder/relocate/StandardHandler.java
```diff
@@ -75,7 +76,10 @@ public HandlingMode applicability(Item item) {
     @Override public List<? extends ItemGroup<?>> validDestinations(Item item) {
         List<DirectlyModifiableTopLevelItemGroup> result = new ArrayList<DirectlyModifiableTopLevelItemGroup>();
         Jenkins instance = Jenkins.getActiveInstance();
-        if (permitted(item, instance)) {
+        // ROOT context is only added in case there is not any item with the same name
+        // But we add it in case the one is there is the item itself and not a different job with the same name
+        // No-op by default
+        if (permitted(item, instance) && (instance.getItem(item.getName()) == null) || instance.getItem(item.getName())==item) {
```
recena: `instance.getItem(item.getName()) == item`

recena:
@fbelzunc I'm going to do some tests and LGTM. I'll merge it.

src/main/java/com/cloudbees/hudson/plugins/folder/relocate/StandardHandler.java
fbelzunc: Good catch ;-)

fbelzunc:
@recena I already addressed your comment. Would you mind to merge? - I am preparing a new easy PR for this plugin so would like to work with latest bits.

recena:
@fbelzunc I recommend you to use the Jira issue ID in you commits `-m "[JENKINS-12345] my message"` or to squash your PR. It is easier when you prepare the changelog for the release notes.

fbelzunc:
ok, thanks. I will do next time.

PR #58, opened 2016-04-19
[JENKINS-34200] Redirect to indexing log after creating or editing a ComputedFolder

recena:
[JENKINS-34200](https://issues.jenkins-ci.org/browse/JENKINS-34200)

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
@@ -81,6 +81,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
 import javax.servlet.ServletException;
 import jenkins.model.Jenkins;
 import jenkins.model.ModelObjectWithChildren;
@@ -707,13 +708,14 @@ public abstract class AbstractFolder<I extends TopLevelItem> extends AbstractIte
     }
 
     /**
-     * Where user will be redirected after creating a {@code AbstractFolder}.
+     * Where user will be redirected after creating or reconfiguring a {@code AbstractFolder}.
      *
      * @return A string that represents the redirect location URL.
      *
      * @see javax.servlet.http.HttpServletResponse#sendRedirect(String)
      */
-    protected String getSuccessfulDestination() {
+    @Restricted(NoExternalUse.class)
+    protected @Nonnull String getSuccessfulDestination() {
         return ".";
     }
 
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright 2015 CloudBees, Inc.
+ * Copyright 2015-2016 CloudBees, Inc.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
@@ -40,9 +40,6 @@ import hudson.model.AllView;
 import hudson.model.Descriptor;
 import hudson.model.HealthReport;
 import hudson.model.Item;
-import static hudson.model.Item.CONFIGURE;
-import static hudson.model.Item.CREATE;
-import static hudson.model.Item.DELETE;
 import hudson.model.ItemGroup;
 import static hudson.model.ItemGroupMixIn.loadChildren;
 import hudson.model.Items;
@@ -705,8 +702,19 @@ public abstract class AbstractFolder<I extends TopLevelItem> extends AbstractIte
             if (namingStrategy.isForceExistingJobs()) {
                 namingStrategy.checkName(name);
             }
-            FormApply.success(".").generateResponse(req, rsp, this);
+            FormApply.success(getSuccessfulDestination()).generateResponse(req, rsp, this);
         }
+    }
+
+    /**
+     * Where user will be redirected after creating a {@code AbstractFolder}.
+     *
+     * @return A string that represents the redirect location URL.
+     *
+     * @see javax.servlet.http.HttpServletResponse#sendRedirect(String)
+     */
+    protected String getSuccessfulDestination() {
+        return ".";
     }
 
     protected void submit(StaplerRequest req, StaplerResponse rsp) throws IOException, ServletException, Descriptor.FormException {}
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
@@ -239,11 +239,7 @@ public abstract class ComputedFolder<I extends TopLevelItem> extends AbstractFol
 
     @Override
     protected @Nonnull String getSuccessfulDestination() {
-        FolderComputation<I> computation = getComputation();
-        if (computation != null) {
-            return computation.getSearchUrl() + "console";
-        }
-        return super.getSuccessfulDestination();
+        return computation.getSearchUrl() + "console";
     }
 
     public Map<TriggerDescriptor,Trigger<?>> getTriggers() {
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
@@ -74,6 +74,7 @@ import org.acegisecurity.Authentication;
 import org.apache.commons.io.FileUtils;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.DoNotUse;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.HttpResponse;
 import org.kohsuke.stapler.HttpResponses;
 import org.kohsuke.stapler.QueryParameter;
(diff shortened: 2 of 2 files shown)


amuniz:
:bee:

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -705,10 +702,21 @@ public void doConfigSubmit(StaplerRequest req, StaplerResponse rsp) throws IOExc
             if (namingStrategy.isForceExistingJobs()) {
                 namingStrategy.checkName(name);
             }
-            FormApply.success(".").generateResponse(req, rsp, this);
+            FormApply.success(getSuccessfulDestination()).generateResponse(req, rsp, this);
         }
     }
 
+    /**
+     * Where user will be redirected after creating a {@code AbstractFolder}.
```
jglick: Or reconfiguring.

jglick: (And if [JENKINS-18681](https://issues.jenkins-ci.org/browse/JENKINS-18681) were implemented, this would be even more misleading, as for regular `Folder`s you would not be sent to `configure` at all during creation.)

jglick:
The suggestion from JENKINS-31601 was to change the redirect for any `ComputedFolder`, which would avoid the need for an API. I think it is reasonable to expect that any time you have a computed folder, if you save its configuration you will want to look at the computation result—especially since if the folder is newly configured, there will probably be no children by the time the redirect is served. (I suppose there is a race condition: if computation is very fast, it might have completed by the time the redirect is served.)

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
recena: @jglick Hence, Do you prefer to address [JENKINS-18681](https://issues.jenkins-ci.org/browse/JENKINS-18681) instead of the proposal in this PR?

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -40,9 +40,6 @@
 import hudson.model.Descriptor;
 import hudson.model.HealthReport;
 import hudson.model.Item;
-import static hudson.model.Item.CONFIGURE;
-import static hudson.model.Item.CREATE;
-import static hudson.model.Item.DELETE;
```
recena: Unrelated with this PR but...

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
```diff
@@ -234,6 +235,15 @@ protected void submit(StaplerRequest req, StaplerResponse rsp) throws IOExceptio
         }
     }
 
+    @Override
+    protected @Nonnull String getSuccessfulDestination() {
+        FolderComputation<I> computation = getComputation();
+        if (computation != null) {
```
jglick: So does this actually work? My guess is that it does not, since `scheduleBuild()` has not even been called by the time this is called inside `super.doConfigSubmit`.

And even if you called `scheduleBuild()` first, that would not suffice, since that merely adds an item to the queue, which may not be scheduled immediately. Typically it will be scheduled very soon, since it is just a flyweight master task; potentially there could be a `QueueTaskDispatcher` or a `ComputedFolder.getCauseOfBlockage` override which prevents that, though I am not aware of any currently. At any rate, you must account for the possibility that the `FolderComputation` has not started instantly.

I suspect that what you need to do is delete the `doConfigSubmit` override, then have this method do something like

```java
Queue.Item item = scheduleBuild2(0);
if (item != null) {
    try {
        FolderComputation<?> computation = item.getFuture().getStartCondition().get(1, TimeUnit.SECONDS);
        return computation.getSearchUrl() + "console";
    } catch (TimeoutException x) {
        // fine, ignore
    } catch (InterruptedException|ExecutionException x) {
        // log?
    }
}
return super.getSuccessfulDestination(); // fallback
```

(Even better would be to check to see if the computation _finishes_ very quickly, and if it does so and produces at least one child, redirect to `.` since we can guess that all went more or less OK and the user is more interested in the children than the log. This would not typically be the case for multibranch projects, which generally make network calls that can take a while, but would often be true for things like [JENKINS-33275](https://issues.jenkins-ci.org/browse/JENKINS-33275).)

jglick:
Note that PR title no longer matches current patch.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
jglick: They are separate RFEs. I am just pointing out that

* the current Javadoc only covers one of the two cases where this will be called creation but not reconfiguration)
* if JENKINS-18681 is implemented, for (non-`Computed`) `Folder`s this method will _not_ be called during creation

recena:
@jglick Works fine but I have to read carefully your comment.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
recena: @jglick Before to do some tests and dive into the source code...

```
@Override
protected @Nonnull String getSuccessfulDestination() {
    FolderComputation<I> computation = getComputation();
    if (computation != null) {
        return ((FolderComputation) computation).getSearchUrl() + "console";
    }
    return super.getSuccessfulDestination();
}
```
This seems it works fine. `computation` is never `null` when this method is invoked because:

1. Adding a new `ComputedFolder`, a  `computation` is created here:

`init()` call to `loadComputation()` and this one is:

```
private synchronized void loadComputation() {
    try {
        FileUtils.forceMkdir(getComputationDir());
    } catch (IOException x) {
        LOGGER.log(Level.WARNING, null, x);
    }
    computation = createComputation(null);
    XmlFile file = computation.getDataFile();
    if (file != null && file.exists()) {
        try {
            file.unmarshal(computation);
        } catch (IOException e) {
            LOGGER.log(Level.WARNING, "Failed to load " + file, e);
        }
    }
}
```

1. Editing a `ComputedFolder`, `computation` always exists from the XML file.

jglick: It does seem true that `computation` will not be null. However you still need to wait for the computation you _scheduled_ to start. Otherwise you may be rendering an empty page from the initial (no-parent) dummy `FolderComputation` which has no log file. It depends on the exact timing of the first JS call to `progressText` after the browser renders the redirect.

jglick: It is a little trickier than with `Job`/`Run` where there is a permanent identifier for each run, included in the URL, and they are saved for later perusal. `FolderComputations` are simply overwritten currently (though there is an option to save backup files on disk).

jglick:
:bee: since my current reading of the code is that it should work well enough. Might be wise to leave behind a comment explaining why, though.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
recena: This is other subject. IMO, it should be addressed in core, _how Jenkins handles this kind of dynamic content using ajax_.

This is completely unrelated with this issue. We should not condition this redirecction to computation is started (scheduled). What happend if another implementation of `ComputedFolder` wants to override `getSuccessfulDestination` and redirects to Jenkins home? Does it make sense to wait for?

jglick: > What happend if another implementation of `ComputedFolder` wants to override `getSuccessfulDestination` and redirects to Jenkins home? Does it make sense to wait for?

Obviously such an implementation would not want to do any waiting.

recena: @jglick I'll remove the `if` after this https://github.com/jenkinsci/cloudbees-folder-plugin/pull/60 is merged.

recena:
If we use Jenkins core **1.609.1** the following line in `tasks-top-extra.jelly`:

```
<l:task icon="images/24x24/clock.png" href="${rootURL}/${it.url}build?delay=0" post="true" title="${%Run Now}"/>
```

will generate this code in the front-end:

```
<div class="task">
  <script>
    function postRequest_id139(a) {
      new Ajax.Request(a.href);
      hoverNotification('Done.', a.parentNode);
      return false;
    }
  </script>
  <a onclick="postRequest_id139(this)" href="/jenkins/job/test1/build?delay=0" class="task-icon-link"><img src="/jenkins/static/ac775133/images/24x24/clock.png" style="width: 24px; height: 24px; width: 24px; height: 24px; margin: 2px;" class="icon-clock icon-md"></a>&nbsp;<a onclick="postRequest_id139(this)" href="/jenkins/job/test1/build?delay=0" class="task-link">Run Now</a>
</div>
```

This code won't send a valid `POST` request.

This was solved by @olivergondza here https://github.com/jenkinsci/jenkins/commit/baa192499a12ecd3fd819b14fcd9c33cd96433ef

I propose to upgrade the baseline to `1.614`. What do you think @jglick?

recena:
Really the issue mentioned above was solved partially.

I've sent a new PR to Jenkins core https://github.com/jenkinsci/jenkins/pull/2283

jglick:
> I propose to upgrade the baseline to 1.614.

Why? If there is a core fix, then users of a sufficiently new core will get it.

Anyway off topic for this PR.

jglick:
re:bee:

recena:
@jglick As I see it this PR won't work with the declared Jenkins baseline. For that reason, I asked you for upgrading.

jglick:
> this PR won't work with the declared Jenkins baseline

I am not following. This PR is changing a redirect when creating a folder. Your comment was about a `build` link in the sidebar. What is the relationship? What will not work?

recena:
> this PR won't work with the declared Jenkins baseline

My sentence is not clear. I mean that `run` option link won't work with the current Jenkins core declared. My recommendation about upgrading is very related with [this conversation](https://github.com/jenkinsci/workflow-multibranch-plugin/pull/6#issuecomment-214997054).

PR #60, opened 2016-04-20
Wrong paths (hardcoded). 

recena:
I found this bug when I was working here https://github.com/jenkinsci/cloudbees-folder-plugin/pull/58

@reviewbybees

### DIFF
--- a/src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/tasks-top-extra.jelly
+++ b/src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/tasks-top-extra.jelly
@@ -1,7 +1,7 @@
 <!--
 The MIT License
 
-Copyright 2013-2015 CloudBees.
+Copyright 2013-2016 CloudBees, Inc.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
@@ -27,10 +27,10 @@ THE SOFTWARE.
 
     <j:set var="computation" value="${it.computation}"/>
 
-    <l:task icon="images/24x24/notepad.png" href="${computation.url}" title="${computation.displayName}">
-        <j:if test="${h.hasPermission(it,it.BUILD) and it.buildable}">
-            <l:task icon="images/24x24/clock.png" href="${computation.url}build?delay=0" post="true" title="${%Run Now}"/>
+    <l:task icon="images/24x24/notepad.png" href="${rootURL}/${computation.url}" title="${computation.displayName}">
+        <j:if test="${h.hasPermission(it, it.BUILD) and it.buildable}">
+            <l:task icon="images/24x24/clock.png" href="${rootURL}/${it.url}build?delay=0" post="true" title="${%Run Now}"/>
         </j:if>
-        <l:task icon="images/24x24/terminal.png" href="${computation.url}console" title="${%Log}"/>
+        <l:task icon="images/24x24/terminal.png" href="${rootURL}/${computation.url}console" title="${%Log}"/>
     </l:task>
 </j:jelly>
--- a/src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/tasks-top-extra.jelly
+++ b/src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/tasks-top-extra.jelly
@@ -30,10 +30,10 @@ THE SOFTWARE.
     <j:if test="${computation != null}">
         <j:set var="title" value="${computation.displayName}"/>
     </j:if>
-    <l:task icon="images/24x24/notepad.png" href="${url}/computation/" title="${title}">
+    <l:task icon="images/24x24/notepad.png" href="${computation.getUrl()}" title="${title}">
         <j:if test="${h.hasPermission(it,it.BUILD) and it.buildable}">
-            <l:task icon="images/24x24/clock.png" href="${url}/build?delay=0" post="true" title="${%Run Now}"/>
+            <l:task icon="images/24x24/clock.png" href="${computation.getUrl()}build?delay=0" post="true" title="${%Run Now}"/>
         </j:if>
-        <l:task icon="images/24x24/terminal.png" href="${url}/computation/console" title="${%Log}"/>
+        <l:task icon="images/24x24/terminal.png" href="${computation.getUrl()}console" title="${%Log}"/>
     </l:task>
 </j:jelly>
--- a/src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/tasks-top-extra.jelly
+++ b/src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/tasks-top-extra.jelly
@@ -24,16 +24,13 @@ THE SOFTWARE.
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout"
          xmlns:t="/lib/hudson" xmlns:f="/lib/form" xmlns:i="jelly:fmt" xmlns:this="this">
-    <j:set var="url" value="${h.getNearestAncestorUrl(request,it)}"/>
-    <j:set var="title" value="${%Folder Computation}"/> <!-- fallback, usually invisible -->
+
     <j:set var="computation" value="${it.computation}"/>
-    <j:if test="${computation != null}">
-        <j:set var="title" value="${computation.displayName}"/>
-    </j:if>
-    <l:task icon="images/24x24/notepad.png" href="${computation.getUrl()}" title="${title}">
+
+    <l:task icon="images/24x24/notepad.png" href="${computation.url}" title="${computation.displayName}">
         <j:if test="${h.hasPermission(it,it.BUILD) and it.buildable}">
-            <l:task icon="images/24x24/clock.png" href="${computation.getUrl()}build?delay=0" post="true" title="${%Run Now}"/>
+            <l:task icon="images/24x24/clock.png" href="${computation.url}build?delay=0" post="true" title="${%Run Now}"/>
         </j:if>
-        <l:task icon="images/24x24/terminal.png" href="${computation.getUrl()}console" title="${%Log}"/>
+        <l:task icon="images/24x24/terminal.png" href="${computation.url}console" title="${%Log}"/>
     </l:task>
 </j:jelly>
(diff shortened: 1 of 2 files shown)


src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/tasks-top-extra.jelly
```diff
@@ -30,10 +30,10 @@ THE SOFTWARE.
     <j:if test="${computation != null}">
         <j:set var="title" value="${computation.displayName}"/>
     </j:if>
-    <l:task icon="images/24x24/notepad.png" href="${url}/computation/" title="${title}">
+    <l:task icon="images/24x24/notepad.png" href="${computation.getUrl()}" title="${title}">
         <j:if test="${h.hasPermission(it,it.BUILD) and it.buildable}">
-            <l:task icon="images/24x24/clock.png" href="${url}/build?delay=0" post="true" title="${%Run Now}"/>
+            <l:task icon="images/24x24/clock.png" href="${computation.getUrl()}build?delay=0" post="true" title="${%Run Now}"/>
```
jglick: BTW you can also use `${computation.url}`.

src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/tasks-top-extra.jelly
```diff
@@ -30,10 +30,10 @@ THE SOFTWARE.
     <j:if test="${computation != null}">
         <j:set var="title" value="${computation.displayName}"/>
     </j:if>
-    <l:task icon="images/24x24/notepad.png" href="${url}/computation/" title="${title}">
+    <l:task icon="images/24x24/notepad.png" href="${computation.getUrl()}" title="${title}">
```
jglick: :bug: if you are using `computation` then you must enclose all this in the `if` block that checks that `computation != null` (which also means that the `title` local variable is superfluous: you can inline `${computation.displayName}`).

Probably the previous code was producing a link which would be a 404 if clicked. Not sure if you would ever really run into this—from reading the code, it seems `getComputation()` is nonnull after the constructor has completed.

recena: `computation` can not be `null`. cf. https://github.com/jenkinsci/cloudbees-folder-plugin/pull/58#discussion_r60394609

jglick: That seems to be true, so change `computation` and `getComputation` to be `@Nonnull` (with a note that they may be null temporarily inside the constructor, so subclasses beware), simplify `getCauseOfBlockage()` and `getEstimatedDuration()` accordingly, and simplify this view so it does no null check and inlines `${computation.displayName}`.

recena:
@jglick I addressed some changes since `computation` can not be `null`. What do you think?

jglick:
You need `@SuppressFBWarnings` on `NP_NONNULL_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR`.

src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/tasks-top-extra.jelly
```diff
@@ -24,16 +24,13 @@ THE SOFTWARE.
...
     <j:set var="computation" value="${it.computation}"/>
-    <j:if test="${computation != null}">
-        <j:set var="title" value="${computation.displayName}"/>
-    </j:if>
-    <l:task icon="images/24x24/notepad.png" href="${url}/computation/" title="${title}">
+
+    <l:task icon="images/24x24/notepad.png" href="${computation.url}" title="${computation.displayName}">
         <j:if test="${h.hasPermission(it,it.BUILD) and it.buildable}">
-            <l:task icon="images/24x24/clock.png" href="${url}/build?delay=0" post="true" title="${%Run Now}"/>
+            <l:task icon="images/24x24/clock.png" href="${computation.url}build?delay=0" post="true" title="${%Run Now}"/>
```
jglick: :bug: huh?

src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/tasks-top-extra.jelly
```diff
@@ -24,16 +24,13 @@ THE SOFTWARE.
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout"
          xmlns:t="/lib/hudson" xmlns:f="/lib/form" xmlns:i="jelly:fmt" xmlns:this="this">
-    <j:set var="url" value="${h.getNearestAncestorUrl(request,it)}"/>
```
jglick: Why?

src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/tasks-top-extra.jelly
recena: @jglick `computation.url` is equivalent to `computation.getUrl()` and this method returns:

```
public @Nonnull String getUrl() {
    return folder.getUrl() + "computation/";
}
```
I don't know that Jelly var.

jglick: I know what it returns. I am asking why you are replacing a working link with a 404.

recena: @jglick Anyway, let me try it.

jglick: Remember `doBuild` is on `ComputedFolder`, not `FolderComputation`.

jglick: Basically you just need to revert this part of the patch, I think.

recena: OMG! It is true, I saw my mistake while I tried it.

recena:
@jglick Reviewed. I've prefered to use `${rootURL}/${computation.url}` instead of `${url}/${computation.searchUrl}`.

src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/tasks-top-extra.jelly
```diff
@@ -24,16 +24,14 @@ THE SOFTWARE.
...
-        <j:set var="title" value="${computation.displayName}"/>
-    </j:if>
-    <l:task icon="images/24x24/notepad.png" href="${url}/computation/" title="${title}">
+
+    <l:task icon="images/24x24/notepad.png" href="${url}/${computation.searchUrl}" title="${computation.displayName}">
         <j:if test="${h.hasPermission(it,it.BUILD) and it.buildable}">
             <l:task icon="images/24x24/clock.png" href="${url}/build?delay=0" post="true" title="${%Run Now}"/>
         </j:if>
-        <l:task icon="images/24x24/terminal.png" href="${url}/computation/console" title="${%Log}"/>
+        <l:task icon="images/24x24/terminal.png" href="${url}/${computation.searchUrl}console" title="${%Log}"/>
```
recena: `href=console` It would have been enough since the `console` option only is available under `indexing/`.

src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/tasks-top-extra.jelly
```diff
@@ -24,16 +24,13 @@ THE SOFTWARE.
...
-    <l:task icon="images/24x24/notepad.png" href="${url}/computation/" title="${title}">
-        <j:if test="${h.hasPermission(it,it.BUILD) and it.buildable}">
-            <l:task icon="images/24x24/clock.png" href="${url}/build?delay=0" post="true" title="${%Run Now}"/>
+
+    <l:task icon="images/24x24/notepad.png" href="${rootURL}/${computation.url}" title="${computation.displayName}">
+        <j:if test="${h.hasPermission(it, it.BUILD) and it.buildable}">
+            <l:task icon="images/24x24/clock.png" href="${rootURL}/${it.url}build?delay=0" post="true" title="${%Run Now}"/>
         </j:if>
-        <l:task icon="images/24x24/terminal.png" href="${url}/computation/console" title="${%Log}"/>
+        <l:task icon="images/24x24/terminal.png" href="${rootURL}/${computation.url}console" title="${%Log}"/>
```
recena: `href=console` It would have been enough since the console option only is available under `indexing/`.

jglick: Generally providing (server-)absolute URLs is preferable, since it allows this view to be used in various contexts without fear of 404s.

src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/tasks-top-extra.jelly
```diff
@@ -24,16 +24,13 @@ THE SOFTWARE.
...
-    <j:if test="${computation != null}">
-        <j:set var="title" value="${computation.displayName}"/>
-    </j:if>
-    <l:task icon="images/24x24/notepad.png" href="${url}/computation/" title="${title}">
-        <j:if test="${h.hasPermission(it,it.BUILD) and it.buildable}">
-            <l:task icon="images/24x24/clock.png" href="${url}/build?delay=0" post="true" title="${%Run Now}"/>
+
+    <l:task icon="images/24x24/notepad.png" href="${rootURL}/${computation.url}" title="${computation.displayName}">
+        <j:if test="${h.hasPermission(it, it.BUILD) and it.buildable}">
+            <l:task icon="images/24x24/clock.png" href="${rootURL}/${it.url}build?delay=0" post="true" title="${%Run Now}"/>
```
jglick: I think this is also fine (`getNearestAncestorUrl` is normally used to provide a `View`-specific URL, which is irrelevant for something we are just `POST`ing to), though there was no reason to edit these two lines in this PR.

recena: @jglick I thought it would be a nice moment to apply the same criteria.

jglick:
:bee:

PR #61, opened 2016-05-11
[JENKINS-33819] New API to set the OrphanedItemStrategy field

recena:
[JENKINS-33819](https://issues.jenkins-ci.org/browse/JENKINS-33819)

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
@@ -75,7 +75,6 @@ import org.acegisecurity.Authentication;
 import org.apache.commons.io.FileUtils;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.DoNotUse;
-import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.HttpResponse;
 import org.kohsuke.stapler.HttpResponses;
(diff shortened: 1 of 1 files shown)


recena:
Deploying spanshots...

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
```diff
@@ -108,6 +108,12 @@ protected ComputedFolder(ItemGroup parent, String name) {
         init();
     }
 
+    protected ComputedFolder(ItemGroup parent, String name, OrphanedItemStrategy strategy) {
```
jglick: :ant: Works for this one purpose, but more general and flexible to revert the new constructor and instead add `setOrphanedItemStrategy`.

amuniz: `orphanedItemStrategy` is commented as `/*almost final*/`, so I guess we don't want to modify it in a setter (not sure what was the reason to add that comment).

jglick: That comment is just wrong. `submit` sets it.

recena: @jglick To use `setOrphanedItemStrategy` was my first approach this weekend but while a discussed in private with @amuniz, I decided to change it, especially, for `/*almost final*/`.

jglick:
:bee:

PR #62, opened 2016-05-13
Fix for #JENKINS-34781 cleaning up workspaces

michaelneale:
(this is not ready for merge, more a discussion in code). 

Sparked off by a discussion of https://groups.google.com/forum/#!topic/jenkinsci-dev/DZzQgfeW0uI
I opened https://issues.jenkins-ci.org/browse/JENKINS-34781 as a possible way to reduce the amount of detritus buildup of orphaned workspaces - ie clean them up when a branch job is deleted in the case of multibranch. 

(this is an alternative approach to speeding up the workspace cleanup thread).

Possible issues: 
* testing
* this may block the removal of the orphaned job while it clean out
* Is this even the right idea!


jglick:
:-1:, this should be handled in `Job.delete` (or an `ItemListener.onDeleted`), not here.

michaelneale:
@jglick right - so it makes sense for all job types, not just muiltibranch? (and ItemListener implies it could be done in a plugin, yes? or would you think core is better?)

michaelneale:
https://github.com/jenkinsci/jenkins/pull/2356 instead (does it make more sense there?)

PR #63, opened 2016-05-19
[JENKINS-34939] Hold the lock on this folder only while deleting itself, not its children

jglick:
[JENKINS-34939](https://issues.jenkins-ci.org/browse/JENKINS-34939)

@reviewbybees


src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -609,9 +609,12 @@ public void onDeleted(TopLevelItem item) throws IOException {
     }
 
     @Override
-    protected void performDelete() throws IOException, InterruptedException {
+    public void delete() throws IOException, InterruptedException {
+        checkPermission(DELETE);
```
jglick: Basically copying from `AbstractItem`.

christ66:
🐝

PR #64, opened 2016-05-20
Upgrade to credentials-2.0

stephenc:
@reviewbybees

Fix the credentials-2.0 style of store actions and other misc tidy-up for the ew credentials 2.0 api

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
@@ -61,8 +61,10 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.WeakHashMap;
 import jenkins.model.Jenkins;
 import jenkins.model.TransientActionFactory;
+import net.jcip.annotations.GuardedBy;
 import net.sf.json.JSONObject;
 import org.acegisecurity.Authentication;
 import org.acegisecurity.context.SecurityContext;
@@ -83,6 +85,10 @@ public class FolderCredentialsProvider extends CredentialsProvider {
      */
     private static final Set<CredentialsScope> SCOPES =
             Collections.<CredentialsScope>singleton(CredentialsScope.GLOBAL);
+
+    @GuardedBy("self")
+    private static final WeakHashMap<AbstractFolder<?>,FolderCredentialsProperty> emptyProperties =
+            new WeakHashMap<AbstractFolder<?>, FolderCredentialsProperty>();
 
     /**
      * {@inheritDoc}
@@ -151,6 +157,14 @@ public class FolderCredentialsProvider extends CredentialsProvider {
             if (property != null) {
                 return property.getStore();
             }
+            synchronized (emptyProperties) {
+                property = emptyProperties.get(folder);
+                if (property == null) {
+                    property = new FolderCredentialsProperty(folder);
+                    emptyProperties.put(folder, property);
+                }
+            }
+            return property.getStore();
         }
         return null;
     }
@@ -188,6 +202,11 @@ public class FolderCredentialsProvider extends CredentialsProvider {
          * Our store.
          */
         private transient StoreImpl store = new StoreImpl();
+
+        /*package*/ FolderCredentialsProperty(AbstractFolder<?> owner) {
+            setOwner(owner);
+            domainCredentialsMap = DomainCredentials.migrateListToMap(null, null);
+        }
 
         /**
          * Backwards compatibility.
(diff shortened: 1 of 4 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
```diff
@@ -69,17 +61,38 @@
...
 
+    /**
+     * The valid scopes for this store.
+     */
     private static final Set<CredentialsScope> SCOPES =
             Collections.<CredentialsScope>singleton(CredentialsScope.GLOBAL);
 
+    @GuardedBy("self")
+    private static final WeakHashMap<AbstractFolder<?>,FolderCredentialsProperty> emptyProperties =
+            new WeakHashMap<AbstractFolder<?>, FolderCredentialsProperty>();
```
jglick: IIUC this is designed to solve a problem analogous to the proprietary CJP-1758, correct? To wit: merely saving a folder configuration should not add an empty `FolderCredentialsProperty` to the `config.xml`; this should only be created if and when some credentials are actually added to it.

src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
```diff
@@ -424,88 +476,73 @@ private synchronized boolean updateCredentials(@NonNull Domain domain, @NonNull
...
-                } else {
-                    return Collections.emptySet();
-                }
-            }
-
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public AbstractFolderProperty<?> reconfigure(StaplerRequest req, JSONObject form) throws FormException {
```
jglick: Why?

src/main/resources/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider/help.groovy
```diff
@@ -0,0 +1,55 @@
...
+    text(_("Credentials will be available to: "));
+    ul() {
+        li() {
+            text(_("Authentication: "))
+            code() {
+                text(ACL.SYSTEM.name);
+            }
+        }
+        def p = CredentialsProvider.USE_ITEM;
+        while (false && p != null && !p.enabled) {
```
jglick: ??

src/main/resources/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider/help.groovy
```diff
@@ -0,0 +1,55 @@
...
+                text(_("Users with permission: "))
+                code() {
+                    text(p.group.title);
+                    text("/");
+                    text(p.name);
+                }
+            }
+        }
+    }
+    text(_("Credentials will be available all items within the folder"));
```
jglick: :ant: what?

jglick:
:bee: to the extent I can understand this, but there is no context or explanation for what you are doing here.

src/main/resources/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider/help.groovy
stephenc: On the new Manage Jenkins | Configure Credentials | Providers config options, when configuring the enabled providers, this help text will be available.

The help text is generated by Groovy as we need to check which permission is available. If the admin has not enabled Credentials/UseItem then it should follow Item/Configure

src/main/resources/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider/help.groovy
stephenc: Good catch, debug text left in play, I shall remove the `false &&`

src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
stephenc: To keep the property. If we don't do this then the property will be removed on a form resale as there is no form values in the config.jelly

src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
stephenc: Yes, though this solution will save the empty property in some circumstances.

src/main/resources/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider/help.groovy
jglick: I was noting that the sentence is ungrammatical.

jglick:
re:bee:

PR #65, opened 2016-05-25
[JENKINS-32178] Noting why you should not override getUrlChildPrefix

jglick:
Cf. https://github.com/jenkinsci/branch-api-plugin/pull/35.

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
@@ -324,6 +324,10 @@ public abstract class AbstractFolder<I extends TopLevelItem> extends AbstractIte
         return getRootDirFor(child.getName()); // TODO see comment regarding loadChildren and encoding
     }
 
+    /**
+     * It is unwise to override this, lest links to children from nondefault {@link View}s break.
+     * {@inheritDoc}
+     */
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -324,6 +324,10 @@ public File getRootDirFor(I child) {
         return getRootDirFor(child.getName()); // TODO see comment regarding loadChildren and encoding
     }
 
+    /**
+     * It is unwise to override this, lest links to children from nondefault {@link View}s break.
```
stephenc: Of course if the views were written correctly in the first place then it would be ok to override this

stephenc:
:bee:

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
jglick: Perhaps. I suspect a patch to `View.getDynamic` would suffice. But would need to be tested from various angles, and would require a core update.

stephenc: My point is that we should be filing an issue against core and marking this as a TODO to re-encourage once core has the fix

PR #66, opened 2016-06-22
[FIXED JENKINS-36160] The credentials provider was not considering the folder as an itemgroup when accessed as an Item

stephenc:
@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
@@ -149,6 +149,21 @@ public class FolderCredentialsProvider extends CredentialsProvider {
     /**
      * {@inheritDoc}
      */
+    @NonNull
+    @Override
+    public <C extends Credentials> List<C> getCredentials(@NonNull Class<C> type, @NonNull Item item,
+                                                          @Nullable Authentication authentication,
+                                                          @NonNull List<DomainRequirement> domainRequirements) {
+        if (item instanceof AbstractFolder) {
+            // credentials defined in the folder should be available in the context of the folder
+            return getCredentials(type, (ItemGroup) item, authentication, domainRequirements);
+        }
+        return super.getCredentials(type, item, authentication, domainRequirements);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
     @Override
     public CredentialsStore getStore(@CheckForNull ModelObject object) {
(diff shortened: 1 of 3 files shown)


recena:
🐝

amuniz:
:bee:

src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
```diff
@@ -149,6 +149,21 @@
...
+    @NonNull
+    @Override
+    public <C extends Credentials> List<C> getCredentials(@NonNull Class<C> type, @NonNull Item item,
+                                                          @Nullable Authentication authentication,
+                                                          @NonNull List<DomainRequirement> domainRequirements) {
+        if (item instanceof AbstractFolder) {
+            // credentials defined in the folder should be available in the context of the folder
+            return getCredentials(type, (ItemGroup) item, authentication, domainRequirements);
+        }
+        return super.getCredentials(type, item, authentication, domainRequirements);
```
oleg-nenashev: Maybe makes sense to merge results into a single list. Just in case super's behavior changes

stephenc: No

oleg-nenashev: Thanks for the detailed answer!

armfergom:
🐝 

Just in case, I tested it and works as a charm 👍

src/test/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProviderTest.java
```diff
@@ -0,0 +1,133 @@
...
+import org.junit.Test;
+import org.jvnet.hudson.test.JenkinsRule;
+
+import static org.hamcrest.Matchers.hasItem;
+import static org.hamcrest.Matchers.hasProperty;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.notNullValue;
+import static org.junit.Assert.assertThat;
+
+public class FolderCredentialsProviderTest {
```
armfergom: Shouldn't we have `@Issue(JENKINS-36160)` here?

KostyaSha: not critical and better mark tests with annotations because more tests could be added later.

PR #67, opened 2016-07-12
[JENKINS-33622] Get 2.x UI to show on config page

mjdetullio:
DO NOT MERGE -- mostly for demonstration at the moment

Besides the nasty logic to detect version and display different views based on hardcoded `"2.0"`...

Looks good on `ComputedFolder` types since the Orphaned Item Strategy has `f:optionalBlock` or `f:radioBlock`.  Seemed to switch between 1.x UI and 2.x UI just fine based on Jenkins version in use.

More on that:

`config-scrollspy.js` only applies to the config if a form element has the `block-control` CSS class applied to it.

https://github.com/jenkinsci/jenkins/blob/jenkins-2.12/war/src/main/js/widgets/config/tabbar.js#L19

Only `f:optionalBlock` and `f:radioBlock` apply this class.

https://github.com/jenkinsci/jenkins/blob/jenkins-2.12/core/src/main/resources/lib/form/optionalBlock.jelly#L71
https://github.com/jenkinsci/jenkins/blob/jenkins-2.12/core/src/main/resources/lib/form/radioBlock.jelly#L61

The standard `Folder` contains none of those elements, so the new UI is not applied.  This PR forces application by adding that class to a field.  There may be other folder types with this problem (`ComputedFolder` is ok).

If forced to apply the new UI to a Folder, the save/apply buttons are out of place until the DOM is modified by expanding config elements or by resizing the browser window (shown below)

image

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
@@ -61,6 +61,7 @@ import hudson.util.FormApply;
 import hudson.util.FormValidation;
 import hudson.util.Function1;
 import hudson.util.HttpResponses;
+import hudson.util.VersionNumber;
 import hudson.views.DefaultViewsTabBar;
 import hudson.views.ViewsTabBar;
 import java.io.File;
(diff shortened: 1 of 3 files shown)


jglick:
The diff seems fine. I would have to study the issues you refer to in order to understand what the problems are, and probably someone like @tfennelly would need to jump in.

jglick:
Resurrected in #115.

PR #70, opened 2016-10-07
Reject attempts to call ChildObserver.created with something that is not in fact a child of this folder

jglick:
@reviewbybees


svanoort:
🐝

PR #72, opened 2016-10-20
[JENKINS-39136] Move CustomNameJobColumn to cloudbees-folder plugin

stephenc:
See [JENKINS-39136](https://issues.jenkins-ci.org/browse/JENKINS-39136)

Also reworked the configuration UI so that this is at least usable by users who might want a item name column using a custom name

screen shot 2016-10-20 at 12 03 07

@reviewbybees


stephenc:
screen shot 2016-10-20 at 12 19 23

Now with the visual hinting (in this case the message is from a bundle in the folders plugin but will reflect the plugin from which the message came)

src/main/java/com/cloudbees/hudson/plugins/folder/views/CustomNameJobColumn.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    public boolean isPreset() {
+        return StringUtils.isNotBlank(bundle);
+    }
+
+    public String getFrom() {
+        if (origin != null) {
+            return Messages.CustomNameJobColumn_From(
+                    origin.getLongName().replace("Hudson", "Jenkins").replace("hudson", "jenkins"), origin.getUrl());
```
jglick: Huh? Best to add a comment explaining what this is for.

stephenc:
After discussion with @jglick, decision is just to remove the `CustomNameJobColumn` altogether

PR #73, opened 2016-10-24
Cache the folder build health

stephenc:
@reviewbybees 

Also remove unused `help.groovy` that should have been removed in #64


src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -123,6 +129,10 @@
 public abstract class AbstractFolder<I extends TopLevelItem> extends AbstractItem implements TopLevelItem, ItemGroup<I>, ModifiableViewGroup, StaplerFallback, ModelObjectWithChildren, StaplerOverridable {
 
     private static final Logger LOGGER = Logger.getLogger(AbstractFolder.class.getName());
+    private static final Random ENTROPY = new Random();
+    private static final int HEALTH_REPORT_CACHE_REFRESH_MIN = Math.max(10, Math.min(1440, Integer.getInteger(
```
jglick: Would be polite to say

```java
/* 1d */1440
```

src/main/resources/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider/help.groovy
```diff
@@ -1,55 +0,0 @@
...
- *
- * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
- * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
- * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
- * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
- * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
- * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
- * THE SOFTWARE.
- */
-package com.cloudbees.hudson.plugins.folder.properties.FolderCredentialsProvider
```
jglick: What does this have to do with the rest of the change? And why is it being deleted?

stephenc: As mentioned on the PR. This file should have been deleted as part of #64 as it is unused (since the property is no longer visible on the config screen *and* anyway it was in the wrong folder too

stephenc:
Metrics implemented in other plugins can add the type info to say they do not need recursive... if all metrics in a folder are non-recursive, then we get a faster calculation... we assume metrics are recursive by default

The metrics will be cached on average 1h so if a metric depends on something which changes outside of a build (e.g. A metric that is built from say downstream project health... don't know if any) then that could be anything up to 90min stale before it gets a refresh... but such a status would be incredibly risky of infinite loops, so not a realistic concern in my book.

So we are left with most (I argue all) metrics that will only change if the job changes or if there is a new build... which is where the listeners come in and invalidate all the way to the root (there is probably another layer of optimisation that we could do in terms of the calculation starting at the route could build up and pre-populate the child folders as it goes... but I rejected as overly complex and YANGNI!

PR #74, opened 2016-10-24
[JENKINS-39213] Add ability for computed folders to control the views in the folder

stephenc:
@reviewbybees


src/main/java/com/cloudbees/hudson/plugins/folder/views/AbstractFolderViewHolder.java
```diff
@@ -0,0 +1,130 @@
...
+ * </li>
+ * <li>
+ * Where the {@link AbstractFolder} has a fixed set of pre-configured views, the plugin can provide
+ * a custom implementation that returns the fixed set of views.
+ * </li>
+ * </ul>
+ *
+ * @since FIXME
+ */
+public abstract class AbstractFolderViewHolder implements ExtensionPoint {
```
jglick: Why is this an `ExtensionPoint`? IIUC it comes from `newFolderViewHolder`, not an `@Extension`.

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
```diff
@@ -45,27 +45,64 @@ THE SOFTWARE.
 
         <st:include page="configure-details" optional="true"/>
 
-        <j:if test="${it.views.size()>1}">
-          <f:entry title="${%Default View}" field="defaultView">
-            <select class="setting-input" name="primaryView">
-              <j:forEach var="v" items="${it.views}">
+        <j:choose>
```
jglick: Seems like this should all be moved to `DefaultFolderViewHolder/config.jelly`.

src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
```diff
@@ -228,6 +237,78 @@ public TopLevelItem newInstance(ItemGroup parent, String name) {
 
     }
 
+    public static class LockedDownSampleComputedFolder extends SampleComputedFolder {
```
jglick: Were you going to write a test which uses this?

stephenc: Damn I could have sworn I did

src/main/java/com/cloudbees/hudson/plugins/folder/views/AbstractFolderViewHolder.java
stephenc: Because not all `ExtensionPoint`s are `@Extension` but all `@Extension`s are `ExtensionPoint`

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
stephenc: Each concrete folder type only ever allows one view holder... I'd need to make it an `AbstractDescribable` if I wanted to move it to there... and then we end up with a drop down selector that I need to force read-only... just as messy when I tried it

jglick:
> not all `ExtensionPoint`s are `@Extension`

Are there some other examples somewhere? I do not know of any, and Javadoc does not indicate that. Certainly there is no purpose in marking every implementable type in the API an extension point.

jglick:
> we end up with a drop down selector that I need to force read-only

`f:property` inlines configuration without visible change in UI (other than possible change in grouping of sections).

jglick:
So is there a downstream PR yet?

stephenc:
@jglick there is one in preparation... but I need to finesse how the feature gets exposed to scm-api. I'm happy with this abstraction... though making the views read-only currently seems a bit more hacky than I'd like... so rather let's hold merging off until I have finished the scm-api changes

stephenc:
@jglick can we give this one priority of review... I believe this is first in the release chain

jglick:
A link to a downstream PR would be helpful for context.

stephenc:
Still battling with scm-api changes to let me publish the branch-api PR that depends on this

jglick:
Sorry, ran out of time to review this one.

stephenc:
https://github.com/jenkinsci/branch-api-plugin/pull/58 is an example of use, e.g. https://github.com/jenkinsci/branch-api-plugin/pull/58/files#diff-b347159a8a8075d4e700065f8668b804R181 and https://github.com/jenkinsci/branch-api-plugin/pull/58/files#diff-d7c2a1fd649da2a97a79573de31e2884R279

stephenc:
@jglick can we get this one over the line?

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
jglick: > then we end up with a drop down selector that I need to force read-only

`f:property`?

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
```diff
@@ -25,19 +25,44 @@
...
 /**
  * Category of {@link AbstractFolder}.
  * @since 4.11-beta-1
  */
-public abstract class AbstractFolderDescriptor extends TopLevelItemDescriptor {
-    
+public abstract class AbstractFolderDescriptor extends TopLevelItemDescriptor implements IconSpec {
+
+    /**
+     * {@inheritDoc}
```
jglick: You cannot “inherit” documentation on a constructor, because it is not an override.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
```diff
@@ -75,8 +100,74 @@ public String getCategoryId() {
...
+    }
+
+    // TODO remove once baseline 2.0
+    public String getDescription() {
+        Stapler stapler = Stapler.getCurrent();
+        if (stapler != null) {
+            try {
+                WebApp webapp = WebApp.getCurrent();
+                MetaClass meta = webapp.getMetaClass(this);
+                Script s = meta.loadTearOff(JellyClassTearOff.class).findScript("newInstanceDetail");
```
jglick: This is unnecessary. `AbstractFolderDescriptor` subclasses already define `newInstanceDetail.jelly` as

```xml
<j:jelly xmlns:j="jelly:core">
  <j:out value="${it.description}"/>
</j:jelly>
```

and implement a `String getDescription()` method, specifically so that we do not need such hacks.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ChildObserver.java
```diff
@@ -58,4 +63,20 @@
      */
     public abstract void created(I child);
 
+    /**
+     * Returns a copy of the item names that have been observed.
+     *
+     * @return a copy of the item names that have been observed.
+     * @since FIXME
+     */
+    public abstract Set<String> observed();
```
jglick: This is incompatible. And what does it have to do with views?!

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ChildObserver.java
```diff
@@ -58,4 +63,20 @@
...
+
+    /**
+     * Returns a copy of the map of orphaned items keyed by name.
+     *
+     * @return a copy of the map of orphaned items keyed by name.
+     * @since FIXME
+     */
+    public abstract Map<String,I> orphaned();
```
jglick: Ditto.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderComputation.java
```diff
@@ -67,23 +67,35 @@
...
 
+    /** If defined, a number of kB that the event log can grow to before rotation. */
+    @SuppressWarnings("FieldMayBeFinal") // let this be set dynamically by system Groovy script
+    @Nonnull
+    private static int EVENT_LOG_MAX_SIZE = Math.max(1,Integer.getInteger(FolderComputation.class.getName() + ".EVENT_LOG_MAX_SIZE", 150));
+
     /** The associated folder. */
-    private transient final @Nonnull ComputedFolder<I> folder;
+    @Nonnull
+    private transient final ComputedFolder<I> folder;
```
jglick: Please avoid gratuitous reformatting if you want people to review your diffs.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderComputation.java
```diff
@@ -163,11 +181,33 @@ public void save() throws IOException {
...
+    }
+
+    public TaskListener createEventsListener() throws IOException {
+        File eventsFile = getEventsFile();
+        boolean rotate = eventsFile.length() > EVENT_LOG_MAX_SIZE * 1024;
+        OutputStream os;
+        if (BACKUP_LOG_COUNT != null) {
+            os = new ReopenableRotatingFileOutputStream(eventsFile, BACKUP_LOG_COUNT);
+            if (rotate) {
+                ((ReopenableRotatingFileOutputStream) os).rewind();
```
jglick: Will need to be rewritten for external task logging (CC @oleg-nenashev).

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ChildObserver.java
stephenc: This is to do with eventing... I cannot give the example of views without the eventing support and the eventing support requires this in the API.

While strictly true that it breaks backwards compatibility, the `ChildObserver` interface is only implemented within `ComputedFolder` and both of those implementations have the corresponding implementations. The other option would have been to change the contract of `ChildObserver` so that the `created` method is not the entry-point and instead use a different entry point for creating (or make `created` final and have that track the observed entries) either of which would be a bigger breaking change.

For any other people that happen to have used the `ChildObserver` abstract class, prior they would not have had the method available, so they would not have been able to call it.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ChildObserver.java
stephenc: This is to do with eventing... I cannot give the example of views without the eventing support and the eventing support requires this in the API.

While strictly true that it breaks backwards compatibility, the `ChildObserver` interface is only implemented within `ComputedFolder` and both of those implementations have the corresponding implementations. The other option would have been to change the contract of `ChildObserver` so that the `created` method is not the entry-point and instead use a different entry point for creating (or make `created` final and have that track the observed entries) either of which would be a bigger breaking change.

For any other people that happen to have used the `ChildObserver` abstract class, prior they would not have had the method available, so they would not have been able to call it... 

While we do call this method from `updateChildren`, that case is always using `FullReindexChildObserver` which has the method implemented.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
stephenc: This method was required in order to not break descriptions in the case where somebody upgrades folders where the plugin that extends folders has not been updated. 

It is an edge case for sure... you need to have upgraded to Jenkins 2.0 without updating the plugins that extend `Folder`... or you need to have a plugin that has not been updated yet... say https://github.com/jenkinsci/literate-plugin/blob/master/src/main/resources/org/cloudbees/literate/jenkins/LiterateMultibranchProject/newInstanceDetail.jelly which is maybe 45 installations or so... but that is just one example where this is required (and yes I'll probably be updating literate with all the scm api changes and will fix that issue... but until that time this method allows the correct description for literate to be displayed when using the latest folders plugin)

stephenc:
@jglick Comments addressed.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
jglick: > in order to not break descriptions in the case where somebody upgrades folders where the plugin that extends folders has not been updated

I think this was just rejected as an unnecessary defense—the impact is minor and people should have long since upgraded those plugins anyway.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ChildObserver.java
jglick: Never mind—the constructor was always package access, so adding new `abstract` methods is not incompatible.

PR #75, opened 2016-11-08
no-test-jar=false

jglick:
Downstream of https://github.com/jenkinsci/plugin-pom/pull/38.

@reviewbybees


abayer:
:bee:

jglick:
@stephenc **do not release** with this in place; I will try to fix tomorrow.

PR #81, opened 2017-01-09
Add some default content when the events log is empty

stephenc:
@reviewbybees 

screen shot 2017-01-09 at 17 17 15


oleg-nenashev:
re-:bee:

PR #82, opened 2017-01-18
[JENKINS-41124] Provide a mechanism for a computed folder to mangle directory names

stephenc:
See [JENKINS-41124](https://issues.jenkins-ci.org/browse/JENKINS-41124)

To do:

- [x] Some tests of implementing this for computed folder directly here (i.e. not relying on tests in downstream branch-api)
- [x] Create issue to migrate this into core ([JENKINS-41222](https://issues.jenkins-ci.org/browse/JENKINS-41222))
- [x] Add TODO into code with the JIRA issue for migrating the logic into core.
- [x] Some other stuff likely

@reviewbybees @jglick


stephenc:
See https://github.com/jenkinsci/branch-api-plugin/pull/72 for intended usage (also will need to steal some of that PR's tests on migration (with adaptations))

stephenc:
@reviewbybees assuming the tests pass we are ready for review

stephenc:
For anyone wondering what is going on with all the NFC vs NFD test failures.

Prior to 2.0.x of branch api, the Organization folders did not actually record the repository name and just blindly assumed that the repository name would always be a valid filename.

Thankfully, with the two initial implementations of Navigators, we were lucky and the backing systems just banned names that could possibly have alternative NFC vs NFD encodings:

* MultiBranch projects already store the name of the branch in unmolested form within the Branch object, so we only need to worry about `OrganizationFolder`'s children.

* There are only two current `OrganizationFolder` navigators:

  * GitHub which helpfully replaces any non-url safe characters with `-`, so will not be an issue
screen shot 2017-01-20 at 13 09 58

  * BitBucket which also replaces any non-url safe characters with `-` (but without a tooltip giving advance notice), so will also not be an issue.

Any new implementations of `SCMNavigator` should be based off of the new SCM API 2.0 and users should be using Branch API 2.0.x (which does record the unmolested "project name" and attach it to the child objects) so if there are some new `SCMNavigator` implementations that happen to allow for the possibility of returning names with different potential encodings under NFC or NFD rules... that should not be an issue going forward.

(Yes I am being super careful of edge cases that should not even arrive... what else are you going to do... I cannot unsee a problem once I have seen it)

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
```diff
@@ -183,4 +189,53 @@ public boolean isLookAndFeelConfigurable(AbstractFolder<?> folder) {
...
+     *         See <a href="https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx">
+     *         Microsoft's page on "Naming Files, Paths, and Namespaces"</a>
+     *         (What's that you say, "Oh but we only run on Linux", perhaps but users may want to migrate from one OS
+     *         to an other by <strong>moving</strong> their {@code JENKINS_HOME} (or even parts of it) so if you are
+     *         mangling names, be sure to ensure that the mangled name is the same on all OSes
+     *     </li>
+     *     <li><a href="http://unicode.org/reports/tr15/">NFC vs NFD</a> may be a concern as different filesystems
+     *     apply different rules and normalization to the filenames. This is primarily a concern when migrating from
+     *     before having a {@link ChildNameGenerator} to having a {@link ChildNameGenerator} as the migration will
+     *     require inference of the unmangled name from the filesystem, which may or may not match the unmangeld
```
jglick: typo

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
```diff
@@ -183,4 +189,53 @@ public boolean isLookAndFeelConfigurable(AbstractFolder<?> folder) {
...
+     *     <li>Assume that only the characters {@code A-Za-z0-9_.-} are available</li>
+     *     <li>Assume that there are some special reserved names such as {@code .} and {@code ..}</li>
+     *     <li>Assume that there are some problematic names to avoid such as {@code AUX}, {@code CON}, {@code NUL}, etc.
+     *         See <a href="https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx">
+     *         Microsoft's page on "Naming Files, Paths, and Namespaces"</a>
+     *         (What's that you say, "Oh but we only run on Linux", perhaps but users may want to migrate from one OS
+     *         to an other by <strong>moving</strong> their {@code JENKINS_HOME} (or even parts of it) so if you are
+     *         mangling names, be sure to ensure that the mangled name is the same on all OSes
+     *     </li>
+     *     <li><a href="http://unicode.org/reports/tr15/">NFC vs NFD</a> may be a concern as different filesystems
```
jglick: `vs.`

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
```diff
@@ -183,4 +189,53 @@ public boolean isLookAndFeelConfigurable(AbstractFolder<?> folder) {
...
+     *     name from the source of the computation. Now POSIX does not specify how the filesystem is supposed to handle
+     *     encoding of filenames and there can be strange behaviours, e.g.
+     *     <a href="http://stackoverflow.com/a/32663908/1589700">{@link File#listFiles()} is rumoured to always return
+     *     NFC on OS-X</a>
+     *     </li>
+     * </ul>
+     * The {@link ChildNameGenerator} at least allows an {@link AbstractFolder} to apply an on-disk naming strategy
+     * that differs from the names used for the URLs.
+     * <p>
+     * <strong>IF YOU IMPLEMENT A {@link ChildNameGenerator} YOU MUST RETURN A SINGLETON FROM THIS METHOD</strong>
```
jglick: Why does it matter? Is someone comparing object identities here?

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
```diff
@@ -183,4 +189,53 @@ public boolean isLookAndFeelConfigurable(AbstractFolder<?> folder) {
...
+     *
+     * @param <I> A wildcard parameter to assist type matching.
```
jglick: This seems suspicious. There is no type parameter on the descriptor itself. And generally introducing a type parameter used only in a return value indicates a design mistake.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
```diff
@@ -183,4 +189,53 @@ public boolean isLookAndFeelConfigurable(AbstractFolder<?> folder) {
...
+     * @return a <strong>SINGLETON</strong> instance of {@link ChildNameGenerator} for all instanced of the concrete
```
jglick: typo

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -0,0 +1,290 @@
...
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.Map;
+import java.util.WeakHashMap;
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+import jenkins.model.TransientActionFactory;
+
+/**
+ * Provides a way for a {@link ComputedFolder} may need to break the association between the directory names on disk
```
jglick: delete `may need`

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -0,0 +1,290 @@
...
+ * @param <P> the type of {@link AbstractFolder}.
+ * @param <I> the type of {@link TopLevelItem} within the folder.
+ * @since 5.17
+ */
+// TODO migrate this functionality (by changing the base class) into core once baseline Jenkins has JENKINS-41222 merged
+public abstract class ChildNameGenerator<P extends AbstractFolder<I>, I extends TopLevelItem> {
+    /**
+     * The name of the file that contains the actual name of the child item.
+     */
+    public static final String CHILD_NAME_FILE = "name-utf8.txt";
```
jglick: I think just `name.txt` would suffice. There is no reason to make the filename so chatty.

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -0,0 +1,290 @@
...
+            idealNames.remove(trace);
+        }
+    }
+
+    /**
+     * Looks up the {@link Item} to see if we stored the ideal name before invoking the constructor that is having
+     * on-disk side-effects before the object has escaped {@link #beforeCreateItem(AbstractFolder, String, String)}
+     * @param parent
+     * @param item
+     * @return
```
jglick: Please just delete the Javadoc tags if you are not going to give any actual text.

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -0,0 +1,290 @@
...
+        if (itemName == null) return null;
+        synchronized (idealNames) {
+            return idealNames.get(new Trace(parent, itemName));
+        }
+    }
+
+    /**
+     * Infers the {@link Item#getName()} from the {@link Item} instance itself. For a valid implementation, the
+     * {@link ComputedFolder} using this {@link ChildNameGenerator} will be attaching into the {@link Item} the
+     * actual name, typically via a {@link JobProperty} or {@link Action} (beware {@link TransientActionFactory}
```
jglick: I thought the point of the `CHILD_NAME_FILE` was to record this? Why would you also need a `*Property`?

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -0,0 +1,290 @@
...
+     * Challenges include:
+     * <ul>
+     * <li>There are some characters that it would be really bad to return in the item name, such as
+     * {@code "/" / "?" / "#" / "[" / "]" / "\"} as these could end up modifying the effective URL</li>
+     * <li>There are names that it would be bad to return as the item name, such as
+     * {@code "" / "." / ".."} as these could end creating broken effective URLs</li>
+     * </ul>
+     * @param parent the parent within which the item is being loaded.
+     * @param item the partially loaded item (take care what methods you call, the item will not have a reference to
+     *             its parent.
```
jglick: missing rparen

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -0,0 +1,290 @@
...
+     */
+    @CheckForNull
+    public abstract String itemNameFromItem(@Nonnull P parent, @Nonnull I item);
+
+    /**
+     * Infers the directory name in which the {@link Item} instance itself should be stored. For a valid
+     * implementation, the {@link ComputedFolder} using this {@link ChildNameGenerator} will be attaching into the
+     * {@link Item} the actual name, typically via a {@link JobProperty} or {@link Action} (beware
+     * {@link TransientActionFactory} implementations may want to invoke {@link Item#getRootDir()} which will trigger
+     * a stack overflow though, so safer to stick with the {@link JobProperty} or equivalent) . This method's task is to
```
jglick: extra space

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -0,0 +1,290 @@
...
+     * @param item the partially loaded item (take care what methods you call, the item will not have a reference to
+     *             its parent.
+     * @return the name of the item.
+     */
+    @CheckForNull
+    public abstract String itemNameFromItem(@Nonnull P parent, @Nonnull I item);
+
+    /**
+     * Infers the directory name in which the {@link Item} instance itself should be stored. For a valid
+     * implementation, the {@link ComputedFolder} using this {@link ChildNameGenerator} will be attaching into the
```
jglick: Move this text into class Javadoc to avoid duplication from `itemNameFromItem`.

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -0,0 +1,290 @@
...
+     * a stack overflow though, so safer to stick with the {@link JobProperty} or equivalent) . This method's task is to
+     * find that and return the filesystem safe mangled equivalent name stored within or {@code null} if that
+     * information is missing (in which case {@link #dirNameFromLegacy(AbstractFolder, String)}
+     * will be called to try and infer the filesystem safe mangled equivalent name from the disk name that the
+     * {@link Item} is being loaded from.
+     *
+     * Challenges include:
+     * <ul>
+     *     <li>The only really filesystem safe characters are {@code A-Za-z0-9_.-}</li>
+     *     <li>Because of Windows and allowing for users to migrate their Jenkins from unix to windows and vice-versa,
```
jglick: `Windows`

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -0,0 +1,290 @@
...
+     *     some names are reserved names under Windows:
+     *     {@code AUX, COM1, COM2, ..., COM9, CON, LPT1, LPT2, ..., LPT9, NUL, PRN} plus all case variations of these
+     *     names plus the variants where a single {@code .} is appended, you need to map those to something else</li>
+     *     <li>Don't make the filenames too long. Try to keep them under 32 characters. If you can go smaller, even
+     *     better.</li>
+     *     <li>Get it right first time</li>
```
jglick: the

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -0,0 +1,290 @@
...
+     * </ul>
+     *
+     * @param parent the parent within which the item is being loaded.
+     * @param item   the partially loaded item (take care what methods you call, the item will not have a reference to
+     *               its parent.
```
jglick: missing rparen

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -0,0 +1,290 @@
...
+     * <li>Previously the name may have been over-encoded with {@link Util#rawEncode(String)} so you may need to
+     * decode it first</li>
+     * <li>There are some characters that it would be really bad to return in the item name, such as
+     * {@code "/" / "?" / "#" / "[" / "]" / "\"} as these could end up modifying the effective URL</li>
+     * <li>There are names that it would be bad to return as the item name, such as
+     * {@code "" / "." / ".."} as these could end creating broken effective URLs</li>
+     * </ul>
+     * @param parent the parent within which the item is being loaded.
+     * @param legacyDirName the directory name that we are loading an item from.
+     * @return
```
jglick: delete empty tags

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -0,0 +1,290 @@
...
+     */
+    @Nonnull
+    public abstract String itemNameFromLegacy(@Nonnull P parent, @Nonnull String legacyDirName);
+
+    /**
+     * {@link #dirNameFromLegacy(AbstractFolder, String)} could not help, we are loading the item for the first
```
jglick: This _is_ `dirNameFromLegacy`, did you mean something else?

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -0,0 +1,290 @@
...
+     * time since the {@link ChildNameGenerator} was enabled for the parent folder type, this method's mission is
+     * to pretend the {@code legacyDirName} is the "mostly correct" name and turn this into the filesystem safe
+     * mangled equivalent name to use going forward.
+     *
+     * Challenges include:
+     * <ul>
+     *     <li>The only really filesystem safe characters are {@code A-Za-z0-9_.-}</li>
+     *     <li>Because of Windows and allowing for users to migrate their Jenkins from unix to windows and vice-versa,
```
jglick: Typos. And heavily duplicated documentation. Move it to class Javadoc and refer to that from here.

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -0,0 +1,290 @@
...
+     *     {@code AUX, COM1, COM2, ..., COM9, CON, LPT1, LPT2, ..., LPT9, NUL, PRN} plus all case variations of these
+     *     names plus the variants where a single {@code .} is appended, you need to map those to something else</li>
+     *     <li>Don't make the filenames too long. Try to keep them under 32 characters. If you can go smaller, even
+     *     better.</li>
+     *     <li>Get it right first time</li>
+     * </ul>
+     *
+     * @param parent        the parent within which the item is being loaded.
+     * @param legacyDirName the directory name that we are loading an item from.
+     * @return
```
jglick: del empty

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -0,0 +1,290 @@
...
+     */
+    @Nonnull
+    public abstract String dirNameFromLegacy(@Nonnull P parent, @Nonnull String legacyDirName);
+
+    public abstract void recordLegacyName(P parent, I item, String legacyDirName) throws IOException;
```
jglick: What does this do?

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -0,0 +1,290 @@
...
+
+    /**
+     * Traces the creation of a new Item in a folder.
```
jglick: `@link`

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
```diff
@@ -151,7 +151,39 @@ protected final void init() {
...
-        loadComputation();
+        synchronized (this) {
+            computation = createComputation(null);
+        }
+    }
+
+    @Override
+    public void onCreatedFromScratch() {
+        try {
+            FileUtils.forceMkdir(getComputationDir());
```
jglick: How is this related?

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
```diff
@@ -151,7 +151,39 @@ protected final void init() {
...
+        } catch (IOException x) {
+            LOGGER.log(Level.WARNING, null, x);
+        }
+        synchronized (this) {
+            XmlFile file = computation.getDataFile();
+            if (file.exists()) {
+                try {
+                    file.unmarshal(computation);
```
jglick: And this? What does it have to do with item naming?

src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderComputation.java
```diff
@@ -118,6 +119,7 @@ public void run() {
         StreamBuildListener listener;
         try {
             File logFile = getLogFile();
+            FileUtils.forceMkdir(logFile.getParentFile());
```
jglick: How is this related?

src/main/java/com/cloudbees/hudson/plugins/folder/relocate/RelocationAction.java
```diff
@@ -168,6 +171,10 @@ public Object getStaplerFallback() {
 
         @Override
         public Collection<? extends Action> createFor(@Nonnull Item target) {
+            if (target.getParent() instanceof ComputedFolder) {
+                // cannot move from a computed folder
```
jglick: Fine but how is it related? Better put in a separate PR.

src/test/java/com/cloudbees/hudson/plugins/folder/ChildNameGeneratorAltTest.java
```diff
@@ -0,0 +1,549 @@
...
+                "job/child_two/",
+                "job/child%20three/",
+                "job/leanbh%20cu%CC%81ig/",
+                "job/%D1%80%D0%B5%D0%B1%D0%B5%D0%BD%D0%BE%D0%BA%20%D0%BF%D1%8F%D1%82%D1%8C/", // ребенок пять
+                "job/%E5%84%BF%E7%AB%A5%E5%85%AD/", // 儿童六
+                "job/%E1%84%8B%E1%85%A1%E1%84%8B%E1%85%B5%207/", // 아이 7
+                "job/nin%CC%83o%20ocho/"
+        );
+        instance.assertItemDirs(round,
+                "child_on-1ec93354e47959489d1440d",
```
jglick: We could fit more real characters in here if we used Base32 rather than hex.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
```diff
@@ -183,4 +189,53 @@ public boolean isLookAndFeelConfigurable(AbstractFolder<?> folder) {
...
+     * {@link AbstractFolder} class or {@code null} if no name mangling will be performed.
+     * @since 5.17
+     */
+    // TODO figure out how one could un-wind name mangling if one ever wanted to
+    @CheckForNull
+    public <I extends TopLevelItem> ChildNameGenerator<AbstractFolder<I>,I> childNameGenerator() {
+        return null;
```
jglick: Is there not a sensible default implementation, at least for `ComputedFolder`s? I would not want a scary complex implementation to be copied among downstream plugins.

src/test/java/com/cloudbees/hudson/plugins/folder/ChildNameGeneratorAltTest.java
```diff
@@ -0,0 +1,549 @@
...
+        }
+        buf.append('-');
+        buf.append(hash.substring(0, 23));
+        return buf.toString();
+    }
+
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    public static class ComputedFolderImpl extends ComputedFolder<FreeStyleProject> {
+
+        private Set<String> fatalKids = new TreeSet<String>();
```
jglick: Probably do not need all this machinery in this test. (Suppose this is a copy-paste?)

src/test/java/com/cloudbees/hudson/plugins/folder/ChildNameGeneratorRecTest.java
```diff
@@ -0,0 +1,586 @@
...
+                buf.append('_');
+            }
+        }
+        buf.append('-');
+        buf.append(hash.substring(0, 23));
+        return buf.toString();
+    }
+
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    public static class ComputedFolderImpl extends ComputedFolder<FreeStyleProject> {
```
jglick: Too much duplication!

src/main/java/com/cloudbees/hudson/plugins/folder/relocate/RelocationAction.java
stephenc: I want to ensure that once we allow for name mangling we prevent the risk of moving items

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
stephenc: Previously I had `ChildNameGenerator.idealNames` as an instance scoped field for each generator. Then I switched to a static field and forgot to change this to a recommendation for performance reasons rather than a hard requirement so that `beforeCreateItem` would work

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
stephenc: Yes the branch api implementation should be fine, but i'd rather not move it while fighting fires

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
stephenc: I'd rather be explicit that the contents are always utf-8 encoded

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
stephenc: The point of CHILD_NAME_FILE is to allow an admin to figure out which folder is which. It is write-only from our perspective (though we will use it as second last resort in the event that the `Action` or `Property` is not present)

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
stephenc: Found a bug that the directory was being created too early within the constructor (which with mangled names gives duplicate partially populated directories - including the `config.xml`, which causes problems on reload when the child items are also folders), so when I removed that I needed to ensure the directory is created here.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
stephenc: Same root. Without moving this here `OrganizationFolder` has issues with it's `MultibranchProject` children (didn't affect when we were actually changing the name to the mangled name, but with this change we enable the disconnect and that surfaces the issue)

src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderComputation.java
stephenc: See previous two comments

src/test/java/com/cloudbees/hudson/plugins/folder/ChildNameGeneratorAltTest.java
stephenc: Yes, but this is just a minimal implementation for tests

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -62,14 +62,30 @@
  * for is missing.</li>
  * </ul>
  *
+ * For a valid implementation, the
+ * {@link ComputedFolder} using this {@link ChildNameGenerator} will be attaching into the {@link Item} the
+ * actual name, typically via a {@link JobProperty} or {@link Action} (beware {@link TransientActionFactory}
+ * implementations may want to invoke {@link Item#getRootDir()} which will trigger a stack overflow though, so
+ * safer to stick with the {@link JobProperty} or equivalent). This method's task is to find that
```
jglick: Which method?

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -142,38 +153,30 @@ protected final String idealNameFromItem(@Nonnull P parent, @Nonnull I item) {
...
-     *     <li>The only really filesystem safe characters are {@code A-Za-z0-9_.-}</li>
-     *     <li>Because of Windows and allowing for users to migrate their Jenkins from unix to windows and vice-versa,
-     *     some names are reserved names under Windows:
-     *     {@code AUX, COM1, COM2, ..., COM9, CON, LPT1, LPT2, ..., LPT9, NUL, PRN} plus all case variations of these
-     *     names plus the variants where a single {@code .} is appended, you need to map those to something else</li>
-     *     <li>Don't make the filenames too long. Try to keep them under 32 characters. If you can go smaller, even
-     *     better.</li>
-     *     <li>Get it right first time</li>
+     * <li>The only really filesystem safe characters are {@code A-Za-z0-9_.-}</li>
+     * <li>Because of Windows and allowing for users to migrate their Jenkins from Wnix to Windows and vice-versa,
```
jglick: typo

PR #83, opened 2017-01-23
[JENKINS-41004] Do not report credentials with IDs masked by nearer folders

stephenc:
See [JENKINS-41044](https://issues.jenkins-ci.org/browse/JENKINS-41004)

Relates to https://github.com/jenkinsci/credentials-plugin/pull/77

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
@@ -28,11 +28,14 @@ import com.cloudbees.hudson.plugins.folder.AbstractFolder;
 import com.cloudbees.hudson.plugins.folder.AbstractFolderProperty;
 import com.cloudbees.hudson.plugins.folder.AbstractFolderPropertyDescriptor;
 import com.cloudbees.plugins.credentials.Credentials;
+import com.cloudbees.plugins.credentials.CredentialsMatcher;
 import com.cloudbees.plugins.credentials.CredentialsMatchers;
+import com.cloudbees.plugins.credentials.CredentialsNameProvider;
 import com.cloudbees.plugins.credentials.CredentialsProvider;
 import com.cloudbees.plugins.credentials.CredentialsScope;
 import com.cloudbees.plugins.credentials.CredentialsStore;
 import com.cloudbees.plugins.credentials.CredentialsStoreAction;
+import com.cloudbees.plugins.credentials.common.IdCredentials;
 import com.cloudbees.plugins.credentials.domains.Domain;
 import com.cloudbees.plugins.credentials.domains.DomainCredentials;
 import com.cloudbees.plugins.credentials.domains.DomainRequirement;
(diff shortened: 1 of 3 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
```diff
@@ -119,21 +124,28 @@
...
                                                           @NonNull List<DomainRequirement> domainRequirements) {
-        if (authentication == null) {
-            authentication = ACL.SYSTEM;
-        }
         List<C> result = new ArrayList<C>();
+        Set<String> ids = new HashSet<String>();
         if (ACL.SYSTEM.equals(authentication)) {
             while (itemGroup != null) {
                 if (itemGroup instanceof AbstractFolder) {
+                    if (!((AbstractFolder) itemGroup).getACL().hasPermission(authentication, USE_ITEM)) {
```
rsandell: Well, you just verified three lines up that `authentication` is equal to `ACL.SYSTEM` so this seems redundant to me.

rsandell:
:bug: missing test that verifies a folder credential is chosen when there is a system credential with the same id.

src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
```diff
@@ -119,21 +124,25 @@
...
-                        result.addAll(DomainCredentials.getCredentials(
+                        for (C c : DomainCredentials.getCredentials(
                                 property.getDomainCredentialsMap(),
                                 type,
                                 domainRequirements,
-                                CredentialsMatchers.always()));
+                                CredentialsMatchers.always())) {
+                            if (!(c instanceof IdCredentials) || ids.add(((IdCredentials) c).getId())) {
+                                // if IdCredentials, only add if we havent added already
+                                // if not IdCredentials, always add
```
jglick: Are there any non-`IdCredentials`, really? Should we just deprecate that possibility?

src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
```diff
@@ -164,6 +173,61 @@
...
+                                domainRequirements,
+                                matcher)) {
+                            if (ids.add(c.getId())) {
+                                result.add(CredentialsNameProvider.name(c), c.getId());
+                            }
+                        }
+                    }
+                }
+                if (itemGroup instanceof Item) {
+                    itemGroup = Item.class.cast(itemGroup).getParent();
```
jglick:
```java
itemGroup = ((Item) itemGroup).getParent();
```

src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
stephenc: I've see one implementation - cannot recall where - so I am not going to pull the plug on them until I have the implementation guidelines written and published

stephenc:
@rsandell we should be good now! (assuming the CI server agrees)

PR #85, opened 2017-02-08
[JENKINS-41842] Increase test coverage for computed folder

varyvol:
[JENKINS-41842](https://issues.jenkins-ci.org/browse/JENKINS-41842)

Tests validating:
- No new item can be added manually to a computed folder.
- Trigger causes a computation.

@reviewbybees @stephenc

### DIFF
--- a/src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
+++ b/src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
@@ -62,7 +62,6 @@ import org.junit.Test;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.not;
 import static org.hamcrest.Matchers.notNullValue;
-import static org.hamcrest.Matchers.nullValue;
 import static org.junit.Assert.*;
 import org.junit.Rule;
 import org.jvnet.hudson.test.Issue;
@@ -153,6 +152,30 @@ public class ComputedFolderTest {
         a1.keepLog(false);
         d.recompute(Result.SUCCESS);
         d.assertItemNames(5);
+    }
+
+    @Test
+    public void notAddChildren() throws Exception {
+        JenkinsRule.WebClient client = r.createWebClient();
+        SampleComputedFolder s = r.jenkins.createProject(SampleComputedFolder.class, "s");
+
+        assertThat(client.getPage(s).getByXPath("//a[contains(text(), \"New Item\")]").size(), is(0));
+
+        s.kids.add("A");
+        s.recompute(Result.SUCCESS);
+
+        assertThat(client.getPage(s).getByXPath("//a[contains(text(), \"New Item\")]").size(), is(0));
+    }
+
+    @Test
+    public void runByTrigger() throws Exception {
+        SampleComputedFolder s = r.jenkins.createProject(SampleComputedFolder.class, "s");
+        s.assertItemNames(0);
+
+        s.addTrigger(new PeriodicFolderTrigger("1m"));
+
+        Thread.sleep(65000);
+        s.assertItemNames(1);
     }
 
     /** Verify that running branch projects are not deleted even after an organization folder reindex. */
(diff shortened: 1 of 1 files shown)


kwhetstone:
🐝

src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
```diff
@@ -155,6 +154,30 @@ public void runningBuild() throws Exception {
         d.assertItemNames(5);
     }
 
+    @Test
+    public void notAddChildren() throws Exception {
+        JenkinsRule.WebClient client = r.createWebClient();
+        SampleComputedFolder s = r.jenkins.createProject(SampleComputedFolder.class, "s");
+
+        assertThat(client.getPage(s).getByXPath("//a[contains(text(), \"New Item\")]").size(), is(0));
```
jglick: `assertEquals` suffices.

src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
```diff
@@ -155,6 +154,30 @@ public void runningBuild() throws Exception {
...
+    }
+
+    @Test
+    public void runByTrigger() throws Exception {
+        SampleComputedFolder s = r.jenkins.createProject(SampleComputedFolder.class, "s");
+        s.assertItemNames(0);
+
+        s.addTrigger(new PeriodicFolderTrigger("1m"));
+
+        Thread.sleep(65000);
```
jglick: Ouch, this will make test runs significantly slower. For purposes of the test it would suffice to call `.run()` on the trigger.

varyvol: Yes, but that doesn't verify that FolderCron is actually checking for triggers and calling them. Or do you think to verify that is not worth waiting this extra time for tests to finish?

jglick: I would hope that the periodic calls to `Trigger.run` acc. to `cronTabSpec` are already covered by core tests. `PeriodicFolderTriggerTest` should already be covering `toCrontab` + `toIntervalMillis` + `getInterval`. And since this is not a regression test—unlike JENKINS-33006, there is no known bug you are proving is fixed—it seems to me like a poor tradeoff to make the plugin tests that much slower¹ just for a small bit of added authenticity.

¹Hard to predict by how much—since `concurrency` is on by default, we may just be pausing one “thread” while other test suites run a bit faster on the newly idle CPUs. `ComputedFolderTest` is a suite, run serially, so its individual execution time will be increased by ~70s. If it happens to be the slowest of the last batch of tests to be run, then you will increase `mvn test` time by ~70s, but AFAIK it is random in which order suites are run.

varyvol: Fair enough then. I am changing this to use trigger.run() so that we avoid wasting those around 70 seconds.

src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
```diff
@@ -159,22 +159,25 @@ public void notAddChildren() throws Exception {
...
     public void runByTrigger() throws Exception {
         SampleComputedFolder s = r.jenkins.createProject(SampleComputedFolder.class, "s");
         s.assertItemNames(0);
 
-        s.addTrigger(new PeriodicFolderTrigger("1m"));
+        PeriodicFolderTrigger t = new PeriodicFolderTrigger("1m");
+        s.addTrigger(t);
+        t.run();
+
+        Thread.sleep(2000);
```
jglick: or

```java
r.waitUntilNoActivity();
```

varyvol: Didn't know this one. Thanks Jesse.

PR #86, opened 2017-02-17
[JENKINS-42151] FolderComputation.createEventsListener() should be thread safe

stephenc:
See [JENKINS-42151](https://issues.jenkins-ci.org/browse/JENKINS-42151)

@reviewbybees 

(more PRs to follow downstream)

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/computed/EventOutputStreams.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/EventOutputStreams.java
@@ -137,6 +137,9 @@ public class EventOutputStreams implements Closeable {
         }
         if (content == null || pendingSize >= flushSize || System.nanoTime() - lastFlushNanos > flushIntervalNanos) {
             synchronized (writeLock) {
+                if (!outputFile.canWriteNow()) {
+                    return;
+                }
                 File file = outputFile.get();
                 if (!appendNextOpen || file.length() > rotateSize) {
                     if (fileCount > 0) {
@@ -264,13 +267,22 @@ public class EventOutputStreams implements Closeable {
      * Supplies the current output file destination. We use indirection so that when used from a
      * {@link FolderComputation} the containing {@link Folder} can be moved without keeping the events log open.
      */
-    public interface OutputFile {
+    public static abstract class OutputFile {
         /**
          * Returns the file that output is being sent to.
          *
          * @return the file that output is being sent to.
          */
         @NonNull
-        File get();
+        public abstract File get();
+
+        /**
+         * Returns {@code true} if the output file can be written to now, {@code false} if the write should be delayed.
+         * 
+         * @return {@code true} if the output file can be written to now, {@code false} if the write should be delayed
+         */
+        public boolean canWriteNow(){
+            return true;
+        }
     }
 }
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/computed/EventOutputStreams.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/EventOutputStreams.java
@@ -55,6 +55,7 @@ import org.apache.commons.io.IOUtils;
  * <li>Rotate after it gets too big</li>
  * <li>Not hold the file open indefinitely - ideally only when writing so that it can be moved by other processes</li>
  * </ul>
+ *
  * @since 5.18
  */
 public class EventOutputStreams implements Closeable {
@@ -165,6 +166,7 @@ public class EventOutputStreams implements Closeable {
                         this.pendingSize.addAndGet(-bytes.length);
                         os.write(bytes);
                     }
+                    os.flush();
                 } catch (IOException e) {
                     // ignore
                 } finally {
@@ -177,12 +179,12 @@ public class EventOutputStreams implements Closeable {
 
     /**
      * Gets a new {@link OutputStream}, the caller must close the stream in order to ensure all its output gets written.
+     *
      * @return a new {@link OutputStream}.
      */
     public OutputStream get() {
         return new OutputStream() {
             private final byte[] buf = new byte[1024];
-            private int last = 0;
             private int index = 0;
             private long start = System.nanoTime();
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderComputation.java
```diff
@@ -195,20 +203,25 @@ public File getEventsFile() {
         return new File(folder.getComputationDir(), "events.log");
     }
 
-    public TaskListener createEventsListener() throws IOException {
+    @WithBridgeMethods(TaskListener.class)
+    @Nonnull
+    public synchronized StreamBuildListener createEventsListener() {
```
stephenc: Well events are like a build, but I can make it a less specific type

src/main/java/com/cloudbees/hudson/plugins/folder/computed/EventOutputStreams.java
```diff
@@ -0,0 +1,270 @@
...
+     */
+    private final long flushIntervalNanos;
+    /**
+     * How big to trigger a flush (defaults to 1024)
+     */
+    private final int flushSize;
+    /**
+     * How big to trigger a rotation (defaults to 32k)
+     */
+    private final long rotateSize;
```
stephenc: The version in core will roll on startup and has other issues too, so partial reuse but necessary. I may duplicate this class in core and switch when available

stephenc:
> Seems like it would be a lot simpler to just cache the `FolderComputation` → `TaskListener`.

To do that I would need the following:

1. The task listener would need to not keep the file open (otherwise we block moves) so we need it to be re-openable
2. The task listener would either need to be synchronized or implement buffering like here
3. If buffering, we would need some fancy logic to ensure that content gets flushed appropriately
4. If synchronized, we would critically need to hook up to the relocation logic (we probably need to hook up to that anyway, but the hook can be simply one where we flag a move in progress and thus keep content in the buffer until ready to write)

With this approach, ok the buffering is slightly more complex than syncronized, but because we always flush on close we can avoid more complex logic to ensure content gets flushed when event handling completes and we avoid having to ensure that even handling does not interfere with relocations

jglick:
> we block moves

Of a higher-level, non-computed folder I suppose you mean? I shudder to think what else will break in such a scenario.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/EventOutputStreams.java
```diff
@@ -0,0 +1,289 @@
...
+                                f.renameTo(n);
+                            }
+                        }
+                    } else {
+                        appendNextOpen = false;
+                    }
+                }
+                FileOutputStream os = null;
+                try {
+                    os = new FileOutputStream(file, appendNextOpen);
```
jglick: May as well go up to 1.625.x and use `try`-with-resources.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/EventOutputStreams.java
```diff
@@ -0,0 +1,289 @@
...
+        /**
+         * Returns the file that output is being sent to.
+         *
+         * @return the file that output is being sent to.
+         */
+        @NonNull
+        public abstract File get();
+
+        /**
+         * Returns {@code true} if the output file can be written to now, {@code false} if the write should be delayed.
```
jglick: Seems inherently prone to race conditions, no?

src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderComputation.java
```diff
@@ -195,20 +208,44 @@ public File getEventsFile() {
...
+                    return getEventsFile();
+                }
+
+                @Override
+                public boolean canWriteNow() {
+                    GregorianCalendar timestamp = new GregorianCalendar();
+                    timestamp.setTimeInMillis(System.currentTimeMillis() - 10000L);
+                    Queue.Item probe = new Queue.WaitingItem(timestamp, folder, Collections.<Action>emptyList());
+                    for (QueueTaskDispatcher d: QueueTaskDispatcher.all()) {
+                        if (d.canRun(probe) != null) {
```
jglick: Is this legal to call outside the `Queue` lock?

stephenc:
@jglick are you done reviewing?

src/main/java/com/cloudbees/hudson/plugins/folder/computed/EventOutputStreams.java
stephenc: Yes, but I cannot see another way... plus the move impl we know off will retry if there is a locked file so whoop the buffer saves us ;-)

src/main/java/com/cloudbees/hudson/plugins/folder/computed/EventOutputStreams.java
stephenc: Follow up PR when I bump other things to at least that core

src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderComputation.java
stephenc: Ooooh interesting question there... certainly when the item is one on the queue and not a probe instance that will never go next nor near the queue... we would need the lock... I'll have to mull this over, but my initial feeling is no... if it is not safe I may need to add a mix-in interface and adjust folders-plus

jglick:
> are you done reviewing?

Yeah, I gave an approval on the tip right?

src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderComputation.java
stephenc: Ok! 

1. `Queue.BlockedItem.getCauseOfBlockage()` calls this method (does not use any lock to guard)
2. `Queue.Item.getWhy()` calls `Queue.BlockedItem.getCauseOfBlockage()` (does not use any lock to guard) and is `@Exported`

Thus there are already code paths that execute the `QueueTaskDispatcher.canRun(Queue.Item)` without holding the `Queue` lock.

Thus I deem it no less safe than the current usage.

I have filed [JENKINS-42248](https://issues.jenkins-ci.org/browse/JENKINS-42248) to enhance core with a less race-condition prone way to handle locking the file-system

PR #87, opened 2017-02-27
[JENKINS-37369] Expose the last computation for computed folders

stephenc:
See [JENKINS-37369](https://issues.jenkins-ci.org/browse/JENKINS-37369)

@reviewbybees

Exposes the last computation information (so will only ever be either success or failure until we add computation history support) 

screen shot 2017-02-27 at 13 18 07

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/computed/PseudoRun.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/PseudoRun.java
@@ -108,15 +108,15 @@ public class PseudoRun<I extends TopLevelItem> extends Actionable implements Sta
     }
 
     public HttpResponse doIndex(StaplerRequest request) {
-        return HttpResponses.redirectTo(Jenkins.getActiveInstance().getRootUrlFromRequest() + computation.getUrl() );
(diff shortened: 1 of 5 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/computed/PseudoRun.java
```diff
@@ -0,0 +1,151 @@
...
+        return computation.getLogFile();
+    }
+
+    @Override
+    public Object getStaplerFallback() {
+        return computation;
+    }
+
+    public HttpResponse doIndex(StaplerRequest request) {
+        return HttpResponses.redirectTo(Jenkins.getActiveInstance().getRootUrlFromRequest() + computation.getUrl() );
```
jglick: Just use

```java
return HttpResponses.redirectViaContextPath(computation.getUrl());
```

stephenc:
Given that the build now button is (as of Jenkins 2.50) saying "Schedule a Scan ..." I would expect that the Last XXX columns refer to that scan:

screen shot 2017-03-10 at 11 01 44

stephenc:
Merged resulting from Patrick's comments on the JIRA ticket

PR #89, opened 2017-03-01
[FIXED JENKINS-40921][FIXED JENKINS-33020] Only display triggers that make sense for a folder computation

stephenc:
See [JENKINS-40921](https://issues.jenkins-ci.org/browse/JENKINS-40921) and [JENKINS-33020](https://issues.jenkins-ci.org/browse/JENKINS-33020)

@reviewbybees


src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
```diff
@@ -342,6 +342,17 @@ protected final String getSuccessfulDestination() {
         return triggers.toMap();
     }
 
+    public List<TriggerDescriptor> getTriggerDescriptors() {
```
jglick:
```java
@Restricted(DoNotUse.class)
```

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
```diff
@@ -342,6 +342,17 @@ protected final String getSuccessfulDestination() {
...
+        List<TriggerDescriptor> result = new ArrayList<TriggerDescriptor>();
+        for (TriggerDescriptor d: Trigger.for_(this)) {
+            if (d instanceof TimerTrigger.DescriptorImpl) {
+                continue;
```
jglick: Arguably better done using `DescriptorVisibilityFilter`, so that it would apply automatically to any other UIs offering `Trigger`s here.

src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/configure-details.jelly
```diff
@@ -26,7 +26,9 @@ THE SOFTWARE.
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:f="/lib/form" xmlns:p="/lib/hudson/project">
     <st:include page="configure-entries" optional="true"/>
-    <p:config-trigger/>
+    <f:descriptorList title="${%triggers(it.computation.displayName)}"
+                      descriptors="${it.triggerDescriptors}"
+                      instances="${it.triggers}"/>
```
jglick: I think this only actually works by accident in `config-trigger.jelly`. Better style is to use the `field` attribute.

Either way, you need a `configRoundtrip` test verifying that the databinding code is correct.

stephenc: actually it works because it is *seriously* old-school binding!

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
stephenc: I am assuming that *core* will add `DescriptorVisibilityFilter` support to `Trigger.for_(Item)` in which case this is just a second level filtering

jglick: Not to `for_`, since that operates without context. Rather you would have to call `h.filterDescriptors` when invoking `descriptorList`. Or use a higher-level control which does this automatically, if there is one.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
```diff
@@ -342,7 +342,9 @@ protected final String getSuccessfulDestination() {
         return triggers.toMap();
     }
 
+    @Restricted(NoExternalUse.class) // currently used only by jelly / stapler
     public List<TriggerDescriptor> getTriggerDescriptors() {
+        // TODO remove this once core has support for DescriptorVisibilityFilter on Trigger.for_(Item)
```
jglick: As above, that is a misrepresentation of what I said.

* `for_` should not call `DescriptorVisibilityFilter`
* merely removing this would revert the fix; you would need to _add_ a `DescriptorVisibilityFilter` implementation

stephenc: yes, we would need to add a descriptor visibility filter as part of removing this method

PR #93, opened 2017-03-08
[FIXED JENKINS-35112] Interrupt the computation when deleting

stephenc:
See [JENKINS-35112](https://issues.jenkins-ci.org/browse/JENKINS-35112)

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
@@ -36,6 +36,7 @@ import hudson.model.BuildableItem;
 import hudson.model.Cause;
 import hudson.model.CauseAction;
 import hudson.model.Descriptor;
+import hudson.model.Executor;
 import hudson.model.Item;
 import hudson.model.ItemGroup;
 import hudson.model.Items;
@@ -43,8 +44,10 @@ import hudson.model.Label;
 import hudson.model.Node;
 import hudson.model.Queue;
 import hudson.model.ResourceList;
+import hudson.model.Result;
 import hudson.model.TaskListener;
 import hudson.model.TopLevelItem;
+import hudson.model.User;
 import hudson.model.listeners.ItemListener;
 import hudson.model.queue.CauseOfBlockage;
 import hudson.model.queue.SubTask;
@@ -64,6 +67,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Level;
@@ -71,6 +75,7 @@ import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import javax.servlet.ServletException;
+import jenkins.model.CauseOfInterruption;
 import jenkins.model.Jenkins;
 import jenkins.model.ParameterizedJobMixIn;
 import jenkins.util.TimeDuration;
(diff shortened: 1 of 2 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
```diff
@@ -271,6 +281,31 @@ final void updateChildren(final TaskListener listener) throws IOException, Inter
...
+        checkPermission(DELETE);
+        FolderComputation<I> computation = getComputation();
+        Executor executor = Executor.of(computation);
+        if (executor != null) {
+            LOGGER.log(Level.INFO, "Interrupting {0} in order to delete it", this);
+            executor.interrupt(Result.ABORTED, new CauseOfInterruption.UserInterruption(User.current()));
+            // give it 15 seconds or so to respond to the interrupt
+            long expiration = System.nanoTime() + TimeUnit.SECONDS.toNanos(15);
+            while (executor.isAlive()
+                    && executor.getCurrentExecutable() == computation
```
jglick: Superfluous check I think—it has been a long time since `Executor`s were reused.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
```diff
@@ -271,6 +281,31 @@ final void updateChildren(final TaskListener listener) throws IOException, Inter
...
+                    && expiration - System.nanoTime() > 0L) {
```
jglick: You mean,

```java
expiration > System.nanoTime()
```

? Other than some overflow conditions somewhere around the heat death of the universe, I think that would be easier to read…

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
```diff
@@ -271,6 +281,31 @@ final void updateChildren(final TaskListener listener) throws IOException, Inter
...
+                Thread.sleep(50L);
+            }
+            if (executor.isAlive() && executor.getCurrentExecutable() == computation) {
+                LOGGER.log(Level.WARNING, "Interrupted {0} in order to delete it, but it has not stopped yet", this);
```
jglick: Would it be more appropriate to just throw an `AbortException` here?

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
stephenc: Future proof in case we ever revert back

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
stephenc: System.nanoTime has an arbitrary start that is allowed to be anywhere. So it can wrap at any time. You basically have to do all sums by subtraction to ensure correct handling during roll-over

stephenc: By way of example, `System.nanoTime()` could return `Long.MAX_VALUE-5000000000L` for the first call... so `expiration` would be `Long.MAX_VALUE+10000000000L`, i.e. negative, after even 1 second then `Long.MAX_VALUE+10000000000L < Long.MAX_VALUE-4000000000L` so `expiration > System.nanoTime()` will be false and we would not wait.

By doing the subtraction comparison, we roll back around and get `Long.MAX_VALUE+10000000000L - Long.MAX_VALUE-4000000000L == 6000000000L > 0`

It would be simpler to do the waiting based on `System.currentTimeMillis()` but that is subject to adjustments so is not safe for measuring elapsed time

jglick: Yes, Javadoc does warn about that theoretical possibility.

stephenc: More than theoretical as nanoTime is based on the system counters which have an arbitrary initial value on at least some reasonable set of supported platforms.

PR #98, opened 2017-03-16
[FIXED JENKINS-42680] Prevent infinite loop when a Folder is a child of an AbstractFolder

stephenc:
See [JENKINS-42680](https://issues.jenkins-ci.org/browse/JENKINS-42680)

Based on tests developed by @varyvol  

@reviewbybees

Supercedes and closes #97


src/main/java/com/cloudbees/hudson/plugins/folder/relocate/StandardHandler.java
```diff
@@ -73,7 +77,54 @@ public HandlingMode applicability(Item item) {
         return Items.move((I) item, destination);
     }
 
-    @Override public List<? extends ItemGroup<?>> validDestinations(Item item) {
+    public boolean hasValidDestination(Item item) {
```
stephenc: Adding specialized handling for the `!validDestinations(item).isEmpty()` case

src/main/java/com/cloudbees/hudson/plugins/folder/relocate/StandardHandler.java
```diff
@@ -88,42 +139,44 @@ public HandlingMode applicability(Item item) {
                 if (!permitted(item, itemGroup)) {
                     continue;
                 }
-                ItemGroup<?> p = itemGroup;
-                if (p instanceof Item) {
-                    Item i = (Item) p;
```
stephenc: `i == g` don't know why we were casting an `Item` to `ItemGroup` and then checking if it was still an `Item` and then casting back to `Item`... it was an `Item` all along and we had it in the variable `g` already!

src/main/java/com/cloudbees/hudson/plugins/folder/relocate/StandardHandler.java
```diff
@@ -51,7 +52,10 @@
 @Extension(ordinal=-1000) public final class StandardHandler extends RelocationHandler {
 
     public HandlingMode applicability(Item item) {
-        if (item instanceof TopLevelItem && item instanceof AbstractItem && item.getParent() instanceof DirectlyModifiableTopLevelItemGroup && !validDestinations(item).isEmpty()) {
+        if (item instanceof TopLevelItem
+                && item instanceof AbstractItem
+                && item.getParent() instanceof DirectlyModifiableTopLevelItemGroup
```
oleg-nenashev: Not sure if all "pretend to be Folder" ItemGroups implement this interface (e.g. Freestyle MultiBranch, Matrix, etc.). But not a part of this change

src/main/java/com/cloudbees/hudson/plugins/folder/relocate/StandardHandler.java
```diff
@@ -88,42 +139,44 @@ public HandlingMode applicability(Item item) {
...
+                }
+                // Cannot move an item into a Folder if there is already an item with the same name
+                if (g instanceof Folder) {
+                    // NOTE: test for Folder not AbstractFolder as Folder is mutable by users
+                    Folder folder = (Folder) g;
+                    if (folder.getItem(item.getName()) != null) {
+                        continue;
                     }
-                    // Cannot move an item into a Folder if there is already an item with the same name
-                    if (i instanceof Folder) {
```
stephenc: This was the infinite loop, namely assuming that a `Folder` would only ever have parents that were also `Folder` or `Jenkins`

src/main/java/com/cloudbees/hudson/plugins/folder/relocate/StandardHandler.java
oleg-nenashev: This method will work correctly IFF it is invoked by user with Global Access (SYSTEM & Co). As designed?

src/main/java/com/cloudbees/hudson/plugins/folder/relocate/StandardHandler.java
```diff
@@ -73,7 +77,54 @@ public HandlingMode applicability(Item item) {
...
     }
 
-    @Override public List<? extends ItemGroup<?>> validDestinations(Item item) {
+    public boolean hasValidDestination(Item item) {
+        Jenkins instance = Jenkins.getActiveInstance();
+        if (permitted(item, instance) && instance.getItem(item.getName()) == null) {
+            // we can move to the root if there is none with the same name.
+            return true;
+        }
+        ITEM: for (Item g : instance.getAllItems()) {
```
oleg-nenashev: Ideally we need an iterator-alike method in order to avoid permission checks of all items just to exit on the first "true"

src/main/java/com/cloudbees/hudson/plugins/folder/relocate/StandardHandler.java
stephenc: Correct, this is whether we display the action *for the current user*

src/main/java/com/cloudbees/hudson/plugins/folder/relocate/StandardHandler.java
stephenc: Yep when baseline core is new enough we can switch to iterator

src/main/java/com/cloudbees/hudson/plugins/folder/relocate/StandardHandler.java
stephenc: If you are a valid target for moving into then you must implement `Does BCT support package annotations BTW?`

src/main/java/com/cloudbees/hudson/plugins/folder/relocate/StandardHandler.java
stephenc: Noted in d526244e55d9ae8f84d8b3cfc047d90a8a2ecf02

PR #100, opened 2017-07-13
[JENKINS-45501] cloudbees-folder is failing in the PCT against cores 2.32, 2.42 and 2.60

raul-arabaolaza:
[JENKINS-45501](https://issues.jenkins-ci.org/browse/JENKINS-45501)

Various fixes for `pom.xml` and `FolderTest` including:
* Parent versionand test dependency bump
* Now the copy operations is performed via HttpRequest instead of trying to interact with the GUI
* Workaround for JENKINS-9774

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.30</version>
+    <version>2.29</version>
     <relativePath />
   </parent>
 
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.29</version>
+    <version>2.30</version>
     <relativePath />
   </parent>
 
@@ -20,7 +20,6 @@
   <url>https://wiki.jenkins-ci.org/display/JENKINS/CloudBees+Folders+Plugin</url>
   <properties>
     <jenkins.version>1.642.3</jenkins.version>
-    <jenkins-test-harness.version>2.18</jenkins-test-harness.version>
     <no-test-jar>false</no-test-jar>
   </properties>
   <licenses>
@@ -60,9 +59,8 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>matrix-auth</artifactId>
-      <version>1.3</version>
+      <version>1.6</version>
       <scope>test</scope>
-      <type>jar</type>
     </dependency>
   </dependencies>
 
--- a/src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
+++ b/src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
@@ -47,7 +47,7 @@ import hudson.views.BuildButtonColumn;
 import hudson.views.JobColumn;
 import java.io.IOException;
 import java.net.URL;
-import java.text.MessageFormat;
+import java.net.URLEncoder;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -151,9 +151,8 @@ public class FolderTest {
         // Taken from https://github.com/jenkinsci/jenkins/blob/80aa2c8e4093df270193402c3933f3f1f16271da/test/src/test/java/hudson/jobs/CreateItemTest.java#L68
         r.jenkins.setCrumbIssuer(null);
 
-        URL apiURL = new URL(MessageFormat.format(
-                "{0}/{1}createItem?mode=copy&from={2}&name={3}",
-                r.jenkins.getRootUrl().toString(), f.getUrl().toString(), fromName, toName));
+        URL apiURL = new URL(
+                r.jenkins.getRootUrl().toString() + "/" + f.getUrl().toString() + "createItem?mode=copy&from=" + URLEncoder.encode(fromName, "UTF-8") + "&name=" + URLEncoder.encode(toName, "UTF-8"));
 
         WebRequest request = new WebRequest(apiURL, HttpMethod.POST);
         request.setEncodingType(null);
--- a/src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
+++ b/src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
@@ -26,10 +26,9 @@ package com.cloudbees.hudson.plugins.folder;
 
 import com.cloudbees.hudson.plugins.folder.properties.FolderCredentialsProvider;
 import com.cloudbees.plugins.credentials.domains.DomainCredentials;
-import com.gargoylesoftware.htmlunit.html.HtmlForm;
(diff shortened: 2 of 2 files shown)


src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
```diff
@@ -137,22 +138,27 @@
...
-        fe.focus();
-        fe.type(fromName);
-        r.submit(page.getFormByName("createItem"));
+    private void copyViaHttp(Folder f, JenkinsRule.WebClient wc, String fromName, String toName) throws Exception {
+        // Taken from https://github.com/jenkinsci/jenkins/blob/80aa2c8e4093df270193402c3933f3f1f16271da/test/src/test/java/hudson/jobs/CreateItemTest.java#L68
+        r.jenkins.setCrumbIssuer(null);
+
+        URL apiURL = new URL(MessageFormat.format(
+                "{0}createItem?mode=copy&from={1}&name={2}",
+                r.jenkins.getRootUrl() + "/" + f.getUrl().toString(), fromName, toName));
```
varyvol: Why not something like "{0}/{1}create..."? Seems weird formatting a String and then format one of the parameters too.

src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
```diff
@@ -137,22 +138,27 @@
         JenkinsRule.WebClient wc = r.createWebClient();
 
         // "foo" should copy "child"
-        copyFromGUI(f, wc, "foo", "xyz");
+        copyViaHttp(f, wc, "foo", "xyz");
```
varyvol: Code looks good, but was this copied by GUI to test the GUI itself or just as something necessary to test anything else?

raul-arabaolaza: AFAIU the test is testing that referring to a job via absolute path or relative '/foo' and `foo` works. The way to do that, via GUI or via and HTTP request, is IMHO irrelevant to the test. In fact the GUI at the end makes the same exact request, so this change is just avoiding an unnecesary and flaky step

src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
raul-arabaolaza: You are right...

src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
varyvol: Fair enough then!

varyvol:
🐝

src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
```diff
@@ -137,22 +138,27 @@
...
-        ((HtmlInput)page.getElementById("name")).type(toName);
-        HtmlInput fe = (HtmlInput) page.getElementById("from");
-        fe.focus();
-        fe.type(fromName);
-        r.submit(page.getFormByName("createItem"));
+    private void copyViaHttp(Folder f, JenkinsRule.WebClient wc, String fromName, String toName) throws Exception {
+        // Taken from https://github.com/jenkinsci/jenkins/blob/80aa2c8e4093df270193402c3933f3f1f16271da/test/src/test/java/hudson/jobs/CreateItemTest.java#L68
+        r.jenkins.setCrumbIssuer(null);
+
+        URL apiURL = new URL(MessageFormat.format(
```
stephenc: :bug: this is not a safe way to construct URLs. Just use string concatenation and encode as appropriate, e.g.

```
new URL(r.getURL().toExternalForm()+f.getUrl()+"?mode=copy&from="+URLEncoder.encode(fromName, "UTF-8")+"&name="+URLEncoder.encode(toName, "UTF-8"))
```

*REMEMBER* Use `URLEncoder.encode(...)` for the _query parameter values_ but use `Util.rawEncode(...)` for the _path segments_... and if the path segments have embedded '/' then you need to figure out whether to split and rejoin on '/' or to have those encoded by `Util.rawEncode(...)` as that is designed for only a single path segment

raul-arabaolaza: @stephenc AFAIU encoding is handled by HtmlUnit [here](https://github.com/jenkinsci/htmlunit/blob/master/src/main/java/com/gargoylesoftware/htmlunit/WebClient.java#L1443) is that encoding wrong or is there something else here? Just trying to understand...

stephenc: You are creating a `new URL(String)` there is no HTMLUnit involved in this constructor call at all. Using `MessageFormat` is absolutely certain to be incorrect as it is designed to generate human readable strings not valid URIs

raul-arabaolaza:
@stephenc I believe now I am constructing the URL properly, IIUC I do not need to use `Util.rawEncode` in this concrete case because there are no path segments that can have strange characters, but I will keep it on my mind for future uses. Thanks!

stephenc:
@raul-arabaolaza yes I had mentioned `Util.rawEncode` _because_ everyone makes the mistake of thinking that `URLEncoder.encode()` works for all parts of the URL, each of the different sections of the URL have different escaping rules and `URLEncoder.encode()` only works on the query string values (perhaps keys too, i'd need to double check against the spec)

Any time you see somebody building an URL or trying to format one, chances are that they are doing it wrong... 

You've conflicts to resolve and you are waiting on the tests to start working (i.e. #103) and if you get a green build then it'll be 👍

stephenc:
@raul-arabaolaza can you rebase onto master so we get a CI build?

pom.xml
```diff
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.29</version>
+    <version>2.30</version>
```
stephenc: Given that this will cause https://youtrack.jetbrains.com/issue/IDEA-175538 for people running intelliJ, I would favour keeping this on 2.29 for now

raul-arabaolaza: `2.29` is enough for PCT so ok

raul-arabaolaza:
@stephenc Used `2.29` and all tests green. :)

src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
```diff
@@ -336,7 +341,10 @@ private void copyFromGUI(Folder f, JenkinsRule.WebClient wc, String fromName, St
     @Issue("JENKINS-32487")
     @Test public void shouldAssignPropertyOwnerOnCreationAndReload() throws Exception {
         Folder folder = r.jenkins.createProject(Folder.class, "myFolder");
-        r.jenkins.setAuthorizationStrategy(new ProjectMatrixAuthorizationStrategy());
+        ProjectMatrixAuthorizationStrategy as = new ProjectMatrixAuthorizationStrategy();
```
jglick: Better to delete this test dep and use `MockAuthorizationStrategy`.

PR #102, opened 2017-07-13
[JDK-8170447] This is the only {@link} to an annotation I could find

stephenc:
See [JDK-8170447](https://bugs.openjdk.java.net/browse/JDK-8170447), if this build passes then the JDK bug is the culprit

@reviewbybees

### DIFF
--- a/src/main/resources/com/cloudbees/hudson/plugins/folder/computed/FolderComputation/consoleFull.jelly
+++ b/src/main/resources/com/cloudbees/hudson/plugins/folder/computed/FolderComputation/consoleFull.jelly
@@ -1,0 +1,28 @@
+<!--
+The MIT License
+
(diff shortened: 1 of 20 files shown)


src/main/resources/com/cloudbees/hudson/plugins/folder/computed/FolderComputation/consoleFull.jelly
```diff
@@ -0,0 +1,28 @@
+<!--
+The MIT License
+
+Copyright (c) 2004-2009, Sun Microsystems, Inc., Alan Harder
```
KostyaSha: ?

stephenc: yep it's a direct copy paste from core

PR #106, opened 2017-08-04
[FIXED JENKINS-45984] Fix the crazy logic where 0 was semantically equivalent to 'empty' but actually wasn't

stephenc:
- All this crazy just to match the way LogRotator in core behaves

See [JENKINS-45984](https://issues.jenkins-ci.org/browse/JENKINS-45984)

@reviewbybees


src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
```diff
@@ -268,6 +295,13 @@ public int compare(I i1, I i2) {
                     toRemove.add(item);
                 }
             }
+            if (daysToKeep == -1 && numToKeep == -1) {
```
stephenc: Tests failed if we don't add this

PR #108, opened 2017-09-29
[FIXED JENKINS-47077] Prevent sympathetic harmonization so that interval actually works

stephenc:
@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/computed/PeriodicFolderTrigger.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/PeriodicFolderTrigger.java
@@ -38,7 +38,6 @@ import java.util.Random;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import jenkins.model.Jenkins;
 import org.apache.commons.lang.StringUtils;
 import org.kohsuke.stapler.DataBoundConstructor;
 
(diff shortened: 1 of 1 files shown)


stephenc:
This should only affect the time difference compared not the crontab

jglick:
So you are modifying the return value of `toIntervalMillis`. That will definitely affect the value of `PeriodicFolderTrigger.interval`. It _could_ also affect the return value of `toCrontab` and thus the value of `Trigger.spec`. Unclear from reading whether the changed value would cause one of the threshold calculations to switch to a new value. You are only subtracting 1s and the comparisons start at 5m, yet `doFillIntervalItems` specifically offers exactly 5m as an option (for example), meaning

```java
if (millis < TimeUnit2.MINUTES.toMillis(5)) {
```

could switch from `false` (an exact match) to `true` (off by 1000ms). Thus changing the crontab for the `5m` option from `* * * * *` (every minute) to `H/12 * * * *` (every 5m) IIUC. Intentional?

Saved settings would be affected if you opened and resaved the configuration form.

jglick:
BTW please merge with `master` to recheck the PR build.

stephenc:
A test would need to run for at least 4 minutes to prove the functionality works... not the kind of test I'd like to add here.

jglick:
Understood that testing this would be tricky without some kind of mock clock. Anyway, was

> changing the crontab for the `5m` option from `* * * * *` (every minute) to `H/12 * * * *` (every 5m)

intentional, an accident, or not actually true at all…?

jglick:
ping @stephenc

jglick:
Related to [JENKINS-47077](https://issues.jenkins-ci.org/browse/JENKINS-47077)?

stephenc:
@jglick well ` TimeUnit2.MINUTES.toMillis(5) <  TimeUnit2.MINUTES.toMillis(5) == false` so ` TimeUnit2.MINUTES.toMillis(5) - 1000L <  TimeUnit2.MINUTES.toMillis(5) == false` is not changing anything much there...

stephenc:
@jglick actually I moved the change to only do the delta in the comparison. No change to persistence at all

stephenc:
superseded by #119

PR #110, opened 2017-10-09
[JENKINS-40612] More efficient idiom for hasValidDestination

jglick:
[JENKINS-40612](https://issues.jenkins-ci.org/browse/JENKINS-40612)

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderCron.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderCron.java
@@ -114,11 +114,7 @@ public class FolderCron extends PeriodicWork {
      * @param cal the date to check for.
      */
     public void checkTriggers(final Calendar cal) {
-        Jenkins inst = Jenkins.getInstance();
-        if (inst == null) {
-            return;
-        }
-        for (ComputedFolder<?> p : inst.getAllItems(ComputedFolder.class)) {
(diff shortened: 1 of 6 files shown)


KostyaSha:
So ACLContext is closeable now?

jglick:
> So ACLContext is closeable now?

It has been since it was introduced—that is its whole reason for existence.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderCron.java
```diff
@@ -114,11 +114,7 @@ public void doRun() {
      * @param cal the date to check for.
      */
     public void checkTriggers(final Calendar cal) {
-        Jenkins inst = Jenkins.getInstance();
-        if (inst == null) {
-            return;
-        }
-        for (ComputedFolder<?> p : inst.getAllItems(ComputedFolder.class)) {
+        for (ComputedFolder<?> p : Jenkins.getInstance().getAllItems(ComputedFolder.class)) {
```
stephenc: Don’t we have `allItems` to save sorting the list?

jglick: Yes, we could apply it here too. I did not because it was not the subject of the PR and it did not at first blush seem like terribly performance-sensitive code. Looking more closely I see that this will be run every minute even if triggers are disabled or infrequent, and I see that you changed the same thing in `Trigger.checkTriggers`, so it probably does make sense.

jglick:
@reviewbybees done

PR #112, opened 2017-10-10
Bump up baseline so we can actually use matrix-auth 2.0

jglick:
#110 & #111 did not mix but apparently we are not building merge PR commits (why?!) so I did not notice until after merging both. #111 forced a core baseline update to match `matrix-auth`, and #110 updated baseline past IIRC 2.12 where we start enforcing such things (would also have been caught by `jenkinsVersions` in `buildPlugin`, or PCT). Anyway 6.1.2 is [mostly used by users of 2.60.1+](http://stats.jenkins.io/pluginversions/cloudbees-folder.html) so might as well update it.

@reviewbybees


src/main/java/com/cloudbees/hudson/plugins/folder/ItemDeletion.java
```diff
@@ -1,208 +0,0 @@
...
-import javax.annotation.concurrent.GuardedBy;
-import org.kohsuke.accmod.Restricted;
-import org.kohsuke.accmod.restrictions.NoExternalUse;
-
-/**
- * A {@link Queue.QueueDecisionHandler} that blocks items being deleted from entering the queue.
- *
- * @since TODO
- */
-@Restricted(NoExternalUse.class) // TODO remove once baseline core has JENKINS-35160
```
oleg-nenashev: Assuming that `Restricted` didn't work in plugin POMs for a while, maybe we need to reconfirm there are bogus usages in dependencies

jglick: This was being used as an extension only. The new core baseline contains the identical class.

jglick:
> I would suggest double-checking that all deleted methods have replacements in parent classes.

Yes I tried to confirm all cases.

jglick:
Test failures. Have fix for one, investigating the other.

pom.xml
```diff
@@ -19,7 +19,8 @@
   </description>
   <url>https://wiki.jenkins-ci.org/display/JENKINS/CloudBees+Folders+Plugin</url>
   <properties>
-    <jenkins.version>2.46.3</jenkins.version>
+    <jenkins.version>2.60.1</jenkins.version>
+    <java.level>8</java.level>
```
recena: @jglick Cannot we use a more recent plugin-pom where this is defined?

recena:
Before releasing I suggest to deploy a snapshot and run our automated test suites, included the PCT.

pom.xml
jglick: 2.36 _is_ the most recent. We do not change the defaults for these two properties—plugins are expected to specify values.

jglick:
@reviewbybees done

src/main/java/com/cloudbees/hudson/plugins/folder/views/DefaultFolderViewHolder.java
```diff
@@ -134,19 +130,7 @@ public void setTabBar(ViewsTabBar tabBar) {
...
-                Method migrateLegacyPrimaryAllViewLocalizedName = AllView.class
-                        .getMethod("migrateLegacyPrimaryAllViewLocalizedName", List.class, String.class);
-                primaryView =
-                        (String) migrateLegacyPrimaryAllViewLocalizedName.invoke(null, views, primaryView);
-            } catch (NoSuchMethodException e) {
-                // ignore, Jenkins core does not have JENKINS-38606 fix merged
-            } catch (IllegalAccessException e) {
-                // ignore, Jenkins core does not have JENKINS-38606 fix merged
-            } catch (InvocationTargetException e) {
-                // ignore, Jenkins core does not have JENKINS-38606 fix merged
```
jglick: [JENKINS-47416](https://issues.jenkins-ci.org/browse/JENKINS-47416)

PR #113, opened 2017-10-15
[FIXED JENKINS-47438] Keep track of recent approvals to prevent race conditions

stephenc:
[JENKINS-47438](https://issues.jenkins-ci.org/browse/JENKINS-47438)

@reviewbybees

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ThrottleComputationQueueTaskDispatcher.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ThrottleComputationQueueTaskDispatcher.java
@@ -87,8 +87,10 @@ public class ThrottleComputationQueueTaskDispatcher extends QueueTaskDispatcher 
                 // TODO make the limit configurable
                 return CauseOfBlockage.fromMessage(Messages._ThrottleComputationQueueTaskDispatcher_MaxConcurrentIndexing());
             }
-            synchronized (nonBlocked) {
-                nonBlocked.add(new NonBlockedDetails(now, item));
+            if (!found) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/computed/ThrottleComputationQueueTaskDispatcher.java
```diff
@@ -92,4 +134,14 @@ public int indexingCount(@CheckForNull Node node) {
         return result;
     }
 
+    private static class NonBlockedDetails {
```
jglick: Maybe expressed more easily using a `WeakHashMap<Queue.Item, Long>`?

stephenc: Actually I think there's an even better way to express this

stephenc:
@jglick a customer was facing a critical outage yesterday, test to follow

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ThrottleComputationQueueTaskDispatcher.java
```diff
@@ -92,4 +172,20 @@ public int indexingCount(@CheckForNull Node node) {
         return result;
     }
 
+    /**
+     * Extend {@link WeakReference} to allow association of secondary data without requiring excessive boxing of
+     * {@link Long}s
+     */
+    private static class TimestamppedWeakReference extends WeakReference<Queue.Item> {
```
jglick: typo

PR #79, opened 2016-12-07
[WIP JENKINS-38606 Follow-up] Ensure AllView gets migrated

stephenc:
Downstream of https://github.com/jenkinsci/jenkins/pull/2603

Illustrated usage of the `migrateLegacyPrimaryAllViewLocalizedName` method in a plugin. 
Here we use reflection as we do not want to force a core version bump on folders plugin.

@reviewbybees

PR #114, opened 2017-10-15
[JENKINS-47416] Use a patched copy of the core API for now

jglick:
[JENKINS-47416](https://issues.jenkins-ci.org/browse/JENKINS-47416)

Pending https://github.com/jenkinsci/jenkins/pull/3088, correct a regression in #112 of a sort; the exception was already being thrown as of #79, but swallowed.

@reviewbybees


jglick:
@amuniz the original logging was throwing an exception and thus breaking deserialization.

PR #115, opened 2017-11-30
[JENKINS-33622] Update to Jenkins 2.x configuration page style

jglick:
[JENKINS-33622](https://issues.jenkins-ci.org/browse/JENKINS-33622)

Updates #67 to take advantage of the current 2.x baseline and have a minimal diff.

Tested in Chromium on 2.60.3 with plain folders, multibranch Pipeline folders, and (GitHub) organization folders. All works well with the exception of one case: when you first create a plain folder, the **Save**/**Apply** button bar floats up too high, until you do something to change rendering (Ctrl-+, resize **Description**, expand **Advanced**, etc. etc.). After saving and reconfiguring the problem does not appear. Seems like some kind of core CSS bug, perhaps.

@reviewbybees & @mjdetullio


mjdetullio:
👍

recampbell:
Code review request is to try it to see if it works. @dwnusbaum

jglick:
Maybe @tfennelly knows? Anyway it seems like a fairly minor manifestation of a core bug, compared to the benefits of easy scrolling between organization folder sections for example. @reviewbybees done unless anyone thinks we need to wait for a bottom-sticker fix.

PR #117, opened 2018-02-13
Don't show nested view when clicking apply to rename a folder

dwnusbaum:
Same problem and fix as [JENKINS-17401](https://issues.jenkins-ci.org/browse/JENKINS-17401).

See also jenkinsci/jenkins#3289 which would allow this code to be deleted eventually.

@reviewbybees

PR #118, opened 2018-02-22
[JENKINS-22936] Remove old duplicated rename logic in favor of new rename action

dwnusbaum:
Downstream of jenkinsci/jenkins#3289. Would ideally prevent the need for more PRs like #117.

In the core PR I stated:

> Other Items (e.g. AbstractFolder) will need to modify their sidepanel.

But it looks like that was incorrect, as the rename action shows up in the sidepanel without any modifications. Here is a screenshot:

screen shot 2018-02-22 at 16 05 19

@reviewbybees


dwnusbaum:
Looks like I will need to fix the rename tests.

dwnusbaum:
Cannot add werk-in-progress label but this needs to be updated after changes in the core PR.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -1389,41 +1378,39 @@ protected String getSuccessfulDestination() {
...
         }
-
-        renameTo(newName);
-        rsp.sendRedirect2("../" + newName);
     }
 
     /**
      * Allows a subclass to block renames under dynamic conditions.
      * @return a message if rename should currently be prohibited, or null to allow
+     * @deprecated Override {@link #checkRename} instead.
```
jglick: Make sure you are verifying compatibility with the `ComputedFolder` override: if the call to this from `checkRename` is commented out, some test ought to fail (if not here then in some downstream plugin like `branch-api` or `workflow-multibranch`).

jglick: …and after verifying this, best to update `ComputedFolder` to avoid the deprecation.

dwnusbaum: Ok, I will verify the compatibility and then update `ComputedFolder`.

dwnusbaum: I did not see any downstream tests that appeared to test a rename, so I added a new test here to verify that the override worked before replacing `#renameBlocker` with `#checkRename`. See [ComputedFolderTest#renameWhileComputing](https://github.com/dwnusbaum/cloudbees-folder-plugin/blob/9d31f95c02a305787be69a26687d2a20254ea092/src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java#L486).

src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
```diff
@@ -481,6 +482,23 @@ public void deleteWhileComputing() throws Exception {
...
 
+    @Test
+    public void renameWhileComputing() throws Exception {
+        CoordinatedComputedFolder d = r.jenkins.createProject(CoordinatedComputedFolder.class, "d");
+        d.kids.addAll(Arrays.asList("A", "B"));
+        QueueTaskFuture<Queue.Executable> future = d.scheduleBuild2(0).getFuture();
+        FolderComputation<FreeStyleProject> computation;
+        while (Executor.of((computation = d.getComputation())) == null) {
+            Thread.sleep(50);
+        }
```
Wadeck: I am not sure this will work everytime. In such cases I prefer to foirce the sync my self. You look at the test in https://github.com/jenkinsci-cert/jenkins/pull/157/files#diff-a686e6a07d22a2e89e35edf7c3257853R194. I use the method `waitForStart` that wait for a build to start before going on.

Secondly, I put a semaphore on a custom build in order to determine myself when the build finishes exactly, so you could try to rename it during the job, receive the error then re-try at the end to receive the "success".

dwnusbaum: The `CoordinatedComputedFolder` here uses a `CountDownLatch` in the [method](https://github.com/jenkinsci/cloudbees-folder-plugin/blob/7e6294bc48d4cddc89c129ceaf1ab6d352462425/src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java#L891) that will end up being run by `d.scheduleBuild2(0)`, which will block until [d.onKid("blah")](https://github.com/jenkinsci/cloudbees-folder-plugin/blob/7e6294bc48d4cddc89c129ceaf1ab6d352462425/src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java#L923) is called, so I don't think there is an issue there.

I copied from other tests that actually needed `computation` for what they were testing, but I will change this test to use `waitForStart` instead, and retry checking the name after the computation finishes.

dwnusbaum:
@reviewbybees done

src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
```diff
@@ -87,10 +87,10 @@
 
         String oldName = f.getName();
 
-        HtmlForm cfg = r.createWebClient().getPage(f, "configure").getFormByName("config");
-        cfg.getInputByName("_.name").setValueAttribute("newName");
+        HtmlForm cfg = r.createWebClient().getPage(f, "confirm-rename").getFormByName("config");
```
dwnusbaum: The URL issue went unnoticed because this test constructs the link manually rather than clicking on the actual link that was generated.

PR #119, opened 2018-03-07
Some tweaks on how periodic if not otherwise ran works

stephenc:
- Should fix the "run every 1 minute" bug

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/computed/PeriodicFolderTrigger.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/PeriodicFolderTrigger.java
@@ -34,6 +34,7 @@ import hudson.triggers.Trigger;
 import hudson.triggers.TriggerDescriptor;
 import hudson.util.ListBoxModel;
 import hudson.util.TimeUnit2;
+import java.util.List;
 import java.util.Random;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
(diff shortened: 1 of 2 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/computed/PeriodicFolderTrigger.java
```diff
@@ -90,22 +91,26 @@ public PeriodicFolderTrigger(String interval) throws ANTLRException {
...
-        if (millis < TimeUnit2.MINUTES.toMillis(5)) {
-            return "* * * * *";
+        // we want to ensure the crontab wakes us excessively
+        if (millis <= TimeUnit2.MINUTES.toMillis(5)) {
+            return "* * * * *"; // 0-5min: check every minute
         }
-        if (millis < TimeUnit2.MINUTES.toMillis(10)) {
-            return "H/12 * * * *";
+        if (millis <= TimeUnit2.MINUTES.toMillis(30)) {
+            return "H/12 * * * *"; // 11-30min: check every 5 minutes
```
deyanstoykov: H/12 means "every 12 minutes," not "every 5 minutes"!

fcojfernandez:
@stephenc I'm not really sure if the cron configuration matches the comments added. I suppose you might have made a mistake with the comments and the configuration is right, but I would like you to confirm this point, please. For example, `"H/12 * * * *"; // 11-30min: check every 5 minutes` is for an execution every 12 minutes per my knowledge (and I might be wrong) but the comment says every 5 minutes. You might want to review the `toCrontab` method before I test and merge your PR

stephenc:
Yep, please feel free to update these. The `H/x` syntax is not user friendly using division to specify the multiple!

PR #122, opened 2018-10-01
Migrate into localization-zh-cn-plugin

LinuxSuRen:
See https://github.com/jenkinsci/localization-zh-cn-plugin


LinuxSuRen:
ping @oleg-nenashev

oleg-nenashev:
Not sure what I can help with here @LinuxSuRen . I am not a maintainer of this plugin nowadays

LinuxSuRen:
@oleg-nenashev I don't know that before.

fcojfernandez:
@LinuxSuRen Could you provide more information about why removing these two files? Since they are parts of resource boundles, I don't see any problem in deleting them, but I would like to have more context before merging. I only can guess that they have been migrated to the _localization-zh-cn-plugin_, but at least the link for the commit would be of help.

LinuxSuRen:
@fcojfernandez Here is a proposal about migrating specific language files into one plugin. I'll put all the Chinese localization files into https://github.com/jenkinsci/localization-zh-cn-plugin. See also https://github.com/jenkinsci/jep/blob/master/jep/216/README.adoc.

fcojfernandez:
I would have appreciated if you had posted the links to the migrated files, but it's ok. I'm merging since I've found the following files:
- https://github.com/jenkinsci/localization-zh-cn-plugin/blob/master/cloudbees-folder-plugin/src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure_zh_CN.properties
- https://github.com/jenkinsci/localization-zh-cn-plugin/blob/master/cloudbees-folder-plugin/src/main/resources/com/cloudbees/hudson/plugins/folder/Folder/tasks-create_zh_CN.properties

@LinuxSuRen I think your initiative it's great, but since I do not want to impact other users, I think it's a good idea to provide more details in future PRs. By the way, thanks for your contribution!

LinuxSuRen:
@fcojfernandez Okay, thanks for your advice.

PR #123, opened 2018-10-21
Fix TimeDuration#getTime() call

darxriggs:
This method is deprecated. See quote:
"This method has always returned a time in milliseconds, when various
callers incorrectly assumed seconds. And this spread through the
codebase. So this has been deprecated for clarity in favour of more
explicitly named methods."

According to the Javadoc of scheduleBuild2() the unit of quietPeriod is
seconds whereas getTime() returned seconds. This looks like a bug.
Therefore now using getTimeInSeconds() instead of getTimeInMillis().


darxriggs:
Hi @fcojfernandez and @oleg-nenashev, as some time already passed since this pull-request was created, I just wanted to check if it can be merged?

PR #128, opened 2019-06-05
[JENKINS-48452] Explain the orphaned item strategy.

rsandell:
The description and comments on the ticket suggests that it is only a misunderstanding on what the orphaned item strategy is actually about, so adding a generic explanation that can be changed by implementations for a more context specific explanation.

See [JENKINS-48452](https://issues.jenkins-ci.org/browse/JENKINS-48452).
See downstream PR for screenshots: https://github.com/jenkinsci/branch-api-plugin/pull/157 

* Explain the orphaned item strategy

### Submitter checklist

      * UI Text change only.

@fcojfernandez
@abayer 
@dwnusbaum 
@bitwiseman

### DIFF
--- a/src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/orphanedItemStrategyDescription.properties
+++ b/src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/orphanedItemStrategyDescription.properties
@@ -1,0 +1,2 @@
+blurb=Computed folders can remove items immediately or leave them based on a desired retention strategy. \
+   By default, items will be removed as soon as the folder computation determines they are no longer present.
--- a/src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/configure-details.jelly
+++ b/src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/configure-details.jelly
@@ -30,6 +30,9 @@ THE SOFTWARE.
                       descriptors="${it.triggerDescriptors}"
                       instances="${it.triggers}"/>
     <f:section title="${%Orphaned Item Strategy}">
+        <f:description>
+            <st:include page="orphanedItemStrategyDescription" optional="true"/>
+        </f:description>
         <j:set var="orphanedItemStrategyDescriptors" value="${instance.orphanedItemStrategyDescriptors}"/>
         <f:block>
             <j:choose>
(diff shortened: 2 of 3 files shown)


rsandell:
I have a downstream PR for branch-api on the way as well that shows the UI changes, just waiting for the incremental release.

rsandell:
Aaand I just noticed that incrementals wasn't enabled on this plugin :man_facepalming:

src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/orphanedItemStrategyDescription.properties
```diff
@@ -0,0 +1,2 @@
+blurb=Computed folders can remove items immediately or leave them based on a desired retention strategy. \
```
dwnusbaum: It kind of seems like talking about "computed folders" might be even more confusing to users than having no documentation since it is such a general term. I have a similar concern about the term "folder computation".

Maybe better to not have anything here by default and just let subclasses add documentation as desired using the terms that make sense to users in context? I don't feel too strongly either way, so if you think the default here is helpful then it seems fine to me.

bitwiseman:
```suggestion
blurb=When a folder that generates the items inside itself is refreshed, any item that was present before the refresh but is not preset after is marked as "orphaned".  \
  Orphaned items can be remove immediately or kept based on some retention strategy. \
```

src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/orphanedItemStrategyDescription.properties
```diff
@@ -0,0 +1,2 @@
...
+   By default, items will be removed as soon as the folder computation determines they are no longer present.
```
bitwiseman:
```suggestion
  By default, orphaned items will be removed immediately after they are marked as orphaned, however they may also be kept based on an orphaned item retention strategy.
```

bitwiseman: @dwnusbaum  Hows this wording?  Avoids some of the technical words to speak generally of folders that generate their contents.

src/main/resources/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder/orphanedItemStrategyDescription.properties
rsandell: This is actually the text that is in the official cloudbees documentation :)

dwnusbaum: Ok, if it's already used in official documentation somewhere then it seems best to copy that wording.

PR #130, opened 2019-09-05
Example of using latest plugin-pom and Jenkins bom

markawm:
Please do not merge, this is just an example showing how a plugin would consume the latest [plugin-pom](https://github.com/jenkinsci/plugin-pom/pull/229) & [Jenkins BOM](https://github.com/jenkinsci/jenkins/pull/4150).

c.f. [https://github.com/jenkinsci/cloudbees-folder-plugin/pull/131](https://github.com/jenkinsci/cloudbees-folder-plugin/pull/131)

The jenkins-bom is imported to synchronise dependencies with <jenkins.version> (so long as that is after the first version where the bom is added).
This shows the following benefits to the bom:
* Added dependency `resteasy-jaxrs` has a conflicting (with `jenkins-core`) dependency on `commons-io` - the correct version (2.6) from `jenkins-core` is used without having to specify here.
* Added dependency on `ant` - this will be synchronised to the `jenkins-core` required version without having to keep a version number in this pom synchronised.
* Added dependency on `slf4j-api`, without having to use `${slf4jVersion}` defined in `plugin-pom` which can become out of date as `jenkins-core` moves on.

(the code changes are just quick fixes to compile successfully against 2.190).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>3.2</version>
+    <version>4.0-SNAPSHOT</version>
     <relativePath />
   </parent>
 
@@ -19,7 +19,7 @@
   </description>
   <url>https://wiki.jenkins.io/display/JENKINS/CloudBees+Folders+Plugin</url>
   <properties>
-    <jenkins.version>2.130</jenkins.version>
+    <jenkins.version>2.190-SNAPSHOT</jenkins.version>
     <java.level>8</java.level>
     <no-test-jar>false</no-test-jar>
   </properties>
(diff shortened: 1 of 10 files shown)


pom.xml
```diff
@@ -50,7 +50,32 @@
...
+
   <dependencies>
+    <dependency>
+      <groupId>org.jboss.resteasy</groupId>
+      <artifactId>resteasy-jaxrs</artifactId>
+      <version>4.0.0.Beta5</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.ant</groupId>
+      <artifactId>ant</artifactId>
```
jglick: Why is this dependency needed here?

markawm: It _isn't_ needed. That was the point in the PR text ^^,  that I just added some fake dependencies to illustrate bom usage cf [PR-131](https://github.com/jenkinsci/cloudbees-folder-plugin/pull/130)

jglick: Uh, OK. This plugin is probably not a helpful example of core BOM usage because it does not use any special libraries. The only cases that I know of where you would benefit from importing the core BOM would be a plugin that bundles a library (dep on non-plugin → `WEB-INF/lib/*.jar`) which is _not_ using `pluginFirstClassLoader` and you would otherwise run into conflicts between a transitive dependency of that library and something which is actually accessible via `jenkins-core.jar`. That is why I suggested looking for something that depends on a library that depends on `commons-io` or similar.

markawm: Yeah, true - it was a contrived example with fake dependencies just to illustrate the point.
To your `commons-io` suggestion: 'blueocean-display-url-plugin' depends on `commons-codec` which results on enforcer errors on both slf4j & commons-codec when building against 2.195 so I re-did the example there:
* [With BOM](https://github.com/jenkinsci/blueocean-display-url-plugin/pull/46)
* [Without BOM](https://github.com/jenkinsci/blueocean-display-url-plugin/pull/47)
HTH

PR #135, opened 2019-12-05
Moved documentation from wiki to GitHub

rsandell:
Wiki is now readonly so need to move the docs and releases to github

@fcojfernandez


rsandell:
build fails because windows agent

PR #136, opened 2019-12-12
[JENKINS-60393] Ensure all plugins are loaded to configure the defaults

Dohbedoh:
See [JENKINS-60393](https://issues.jenkins-ci.org/browse/JENKINS-60393): the configuration of the defaults in the constructor of the GlobalConfiguration may cause the loading to fail on startup in some cases

* Ensure all plugin are loaded before setting global defaults

### Submitter checklist

@fcojfernandez @rsandell

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/config/AbstractFolderConfiguration.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/config/AbstractFolderConfiguration.java
@@ -35,15 +35,7 @@ public class AbstractFolderConfiguration extends GlobalConfiguration {
     @DataBoundConstructor
     public AbstractFolderConfiguration() {
         this.load();
-    }
-
-    /**
-     * Auto-configure the default metrics after all plugins have been loaded.
-     */
-    @Initializer(after = InitMilestone.EXTENSIONS_AUGMENTED, before = InitMilestone.JOB_LOADED)
-    public static void autoConfigure() {
-        AbstractFolderConfiguration abstractFolderConfiguration = AbstractFolderConfiguration.get();
-        if (abstractFolderConfiguration.healthMetrics == null) {
+        if(healthMetrics == null) {
             List<FolderHealthMetric> metrics = new ArrayList<>();
             for (FolderHealthMetricDescriptor d : FolderHealthMetricDescriptor.all()) {
                 FolderHealthMetric metric = d.createDefault();
@@ -51,8 +43,8 @@ public class AbstractFolderConfiguration extends GlobalConfiguration {
                     metrics.add(metric);
                 }
             }
-            abstractFolderConfiguration.setHealthMetrics(new DescribableList<FolderHealthMetric,
-                    FolderHealthMetricDescriptor>(abstractFolderConfiguration, metrics));
+            setHealthMetrics(new DescribableList<FolderHealthMetric,
+                    FolderHealthMetricDescriptor>(this, metrics));
         }
     }
 
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/config/AbstractFolderConfiguration.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/config/AbstractFolderConfiguration.java
@@ -7,9 +7,11 @@ import hudson.init.InitMilestone;
 import hudson.init.Initializer;
 import hudson.util.DescribableList;
 import jenkins.model.GlobalConfiguration;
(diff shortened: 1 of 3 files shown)


varyvol:
The change looks good, but given this comes from a bug I think it's worth to have an automated test.

fcojfernandez:
@Dohbedoh I've been doing some testing and I've found weird this:

1. Install from scratch
1. Go to Manage Jenkins > Configure and remove all the `Health Metrics`
1. Save
1. Enter again. The `Health Metrics` are there again

Shouldn't it be removed?

src/main/java/com/cloudbees/hudson/plugins/folder/config/AbstractFolderConfiguration.java
```diff
@@ -30,24 +32,33 @@ public static AbstractFolderConfiguration get() {
 
     @DataBoundConstructor
     public AbstractFolderConfiguration() {
-        load();
-        if (healthMetrics == null) {
+        this.load();
+    }
+
+    /**
+     * Auto-configure the default metrics afetr all plugins have been loaded.
```
rsandell:
```suggestion
     * Auto-configure the default metrics after all plugins have been loaded.
```

src/main/java/com/cloudbees/hudson/plugins/folder/config/AbstractFolderConfiguration.java
```diff
@@ -30,24 +32,33 @@ public static AbstractFolderConfiguration get() {
...
+     */
+    @Initializer(after = InitMilestone.EXTENSIONS_AUGMENTED, before = InitMilestone.JOB_LOADED)
```
rsandell: I don't remember the implications of having both `after` and `before` rules in the same initializer. But I guess you've verified it works as intended?

Dohbedoh: I have tested this yes. I am not sure how to create a test for the initialization though ? 
The real fix is to perform the auto configuration outside the constructor and when plugins are loaded: `after = InitMilestone.EXTENSIONS_AUGMENTED`. I do believe that this is enough. I added the `before = InitMilestone.JOB_LOADED` because I feel this is consistent (the global metrics will be used on Folder's creation)

Dohbedoh:
@fcojfernandez You are right. Apparently when you remove all metrics, no `healthMetrics` attributes is included in the JSON at all. And the bind just does nothing. I need to override the `GlobalConfiguration#configure` method.

Dohbedoh:
@varyvol @rsandell @fcojfernandez any idea how I can write a test for the component initialization problem ?

varyvol:
@Dohbedoh I think if you tried to create a folder, then something was logged? I'd say that's what you need to automate.

src/main/java/com/cloudbees/hudson/plugins/folder/config/AbstractFolderConfiguration.java
```diff
@@ -30,27 +34,47 @@ public static AbstractFolderConfiguration get() {
...
\ No newline at end of file
+
+    @Override
+    public boolean configure(StaplerRequest req, JSONObject json) {
+        if(json.containsKey("healthMetrics")) {
+            req.bindJSON(this, json);
+            this.save();
+        } else {
+            this.setHealthMetrics(Collections.emptyList());
+        }
```
rsandell: Missed `save()` ?

Maybe simpler to 

```java
this.setHealthMetrics(Collections.emptyList());
req.bindJSON(this, json);
this.save();
```

?

rsandell: ah I see now the call to `save()` in `setHealthMetrics`

Dohbedoh: yeah this would `save()` twice

fcojfernandez:
Manual testing seems to be OK. @varyvol 's comment is the only one missing:
>@Dohbedoh I think if you tried to create a folder, then something was logged? I'd say that's what you need to automate.

Dohbedoh:
We would need to cause the circular dependency first though. It cause the component to not be loaded (`AbstractFolderConfiguration#get` thows an `IllegalStateException`). So maybe I can create an anonymous class with an `@Initializer` to cause this ?

PR #101, opened 2017-07-13
[FIXED JENKINS-45322] Computed folder should disable orphaned items

stephenc:
See [JENKINS-45322](https://issues.jenkins-ci.org/browse/JENKINS-45322) (also [INFRA-1262](https://issues.jenkins-ci.org/browse/INFRA-1262) for a real world example where this is needed)

@reviewbybees 

(Additional downstream changes to branch-api to follow in order to prevent event processing on disabled folders)

PR #143, opened 2020-01-29
[JENKINS-60900] Fix enable/disable controls display

Dohbedoh:
See [JENKINS-60900](https://issues.jenkins-ci.org/browse/JENKINS-60900): The controls for the "Disable" feature are not displayed.

* Display Enable / Disable buttons in a Folder's page if it supports it

### Submitter checklist

@fcojfernandez

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
@@ -1230,7 +1230,7 @@ public abstract class AbstractFolder<I extends TopLevelItem> extends AbstractIte
      * @return {@code true} if and only if {@link #setDisabled(boolean)} is implemented
      * @since 6.1.0
      */
-    protected boolean supportsMakeDisabled() {
+    public boolean supportsMakeDisabled() {
         return false;
     }
 
--- a/src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/view-index-top.jelly
+++ b/src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/view-index-top.jelly
@@ -37,10 +37,6 @@ THE SOFTWARE.
       </div>
       <j:out value="${it.description!=null ? app.markupFormatter.translate(it.description) : ''}"/>
   </div>
-  <!-- give actions a chance to contribute summary item -->
-  <j:forEach var="a" items="${it.allActions}">
-    <st:include page="summary" from="${a}" optional="true" it="${a}" />
-  </j:forEach>
   <this:description />
   <j:choose>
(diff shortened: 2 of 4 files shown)


src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/view-index-top.jelly
```diff
@@ -66,4 +62,8 @@ THE SOFTWARE.
       </div>
     </j:otherwise>
   </j:choose>
+  <!-- give actions a chance to contribute summary item -->
```
Dohbedoh: I proposed to move the button up. As an example, with the Ownership plugin the current layout is showing this:

before

By changing the order, we get something like this:

after

More appealing and also closer to what Jobs are showing. Let me know what you think. Maybe this can be improve even more.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -1230,7 +1230,7 @@ protected void setDisabled(boolean disabled) {
      * @return {@code true} if and only if {@link #setDisabled(boolean)} is implemented
      * @since 6.1.0
      */
-    protected boolean supportsMakeDisabled() {
+    public boolean supportsMakeDisabled() {
```
Dohbedoh: Not that it works if we only make the method `ComputedFolder#supportsMakeDisabled` public and keep this one protected but to keep it consistent we should make the parent public too, this would also prevent consumer to make a mistake if they override it.

fcojfernandez:
Thanks for this PR @Dohbedoh ! I will review it by the end of the week

varyvol:
@Dohbedoh it sounds weird to me to be honest. A ComputedFolder in theory should handle things by itself and not by human interaction? I don't know, for example imagine a stalked PR or something like that (just a figured example, not that it's an implemented use case). Maybe that's the reason why it was not done public?

Dohbedoh:
@varyvol Based on the comment https://github.com/jenkinsci/cloudbees-folder-plugin/pull/101#issuecomment-315049172 I understand that the button was intended to be shown. But maybe I am wrong. In addition, a user is already able to disable the ComputedFolder by appending `/disable` to the URL.

varyvol:
@Dohbedoh seems to make sense then, yes. Thanks!

fcojfernandez:
Closing to relaunch the CI

Dohbedoh:
@fcojfernandez Is any action required from me ?

fcojfernandez:
@Dohbedoh It just went out my mind, my apologies!
I closed and re-opened the PR because the failing tests were related to infra issues and later I totally forgot. Sorry again!
For future, please don't wait for so much time. Don't hesitate and ping me

PR #144, opened 2020-03-17
Predicate form of getItems() API

res0nance:
Signed-off-by: Raihaan Shouhell <raihaan.shouhell@autodesk.com>

Downstream of jenkinsci/jenkins#4469

* Entry 1: Developer, predicate form of getItems API
* ...

### Submitter checklist

@fcojfernandez

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
@@ -99,6 +99,7 @@ import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Predicate;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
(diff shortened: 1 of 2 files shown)


fcojfernandez:
thanks for this PR @res0nance  !

This week I don't have availability to review it, but I will have a look next one

src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
```diff
@@ -451,6 +451,14 @@ private void copyViaHttp(Folder f, JenkinsRule.WebClient wc, String fromName, St
         assertFalse(f.hasVisibleItems());
     }
 
+    @Test public void getItemsPredicate() throws IOException {
+        final Folder d = r.jenkins.createProject(Folder.class, "d");
+        final FreeStyleProject p1 = d.createProject(FreeStyleProject.class, "p1");
+        FreeStyleProject p2 = d.createProject(FreeStyleProject.class, "p2");
+        FreeStyleProject c1 = d.createProject(FreeStyleProject.class, "c1");
```
fcojfernandez: just to keep them all the same
```suggestion
        final FreeStyleProject p1 = d.createProject(FreeStyleProject.class, "p1");
        final FreeStyleProject p2 = d.createProject(FreeStyleProject.class, "p2");
        final FreeStyleProject c1 = d.createProject(FreeStyleProject.class, "c1");
```

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -1000,9 +1001,17 @@ public FolderIcon getIconColor() {
     @Exported(name="jobs")
     @Override
     public Collection<I> getItems() {
+        return getItems(item -> true);
+    }
+
+    /**
+     * Gets all the items in this collection in a read-only view that matches supplied Predicate
+     */
+    // TODO: @Override and inherit docs once baseline is above 2.222
```
fcojfernandez: is not needed to update the baseline for this PR?

res0nance: We have to so that we can add the `@Override` annotation but this new function is only available in `2.222` so unless we are comfortable with pushing the baseline so high up for now we can add this code here and add the annotation later.

PR #146, opened 2020-03-26
Fix crontab spec generation

95jonpet:
Overhauled generation of crontab schedules to be more flexible and reflect the actual interval. The previous method did not result in a crontab that matched the interval description. Now, the generation should be more flexible and allow intervals from `1m` (1 minute) to `28d` (28 days, which is the maximum allowed without throwing exceptions when passing it on).

Also added tests related to crontabs and intervals for the `PeriodicFolderTrigger` class.

* Changed generated crontab to reflect specified interval.
* Internal: Added PeriodicFolderTrigger tests.

@fcojfernandez

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/computed/PeriodicFolderTrigger.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/PeriodicFolderTrigger.java
@@ -88,27 +88,31 @@ public class PeriodicFolderTrigger extends Trigger<ComputedFolder<?>> {
      * @return the crontab.
      */
     private static String toCrontab(String interval) {
-        long millis = toIntervalMillis(interval);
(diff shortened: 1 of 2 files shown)


95jonpet:
Is there a way to re-trigger the tests? One temp folder was generated with a space in its name for the Windows build, and that seems to have caused an issue. The Linux build is green. Tested locally on Windows 10 with Maven 3.6.3 and OpenJDK 1.8.0-242.

fcojfernandez:
Closing & reopening the PR to re-launch the CI

varyvol:
I agree it seems risky to change something so widespread, unless there is a clear bug or something very bad is happening from it.

I am also curious: why did you not take into account 's' or 'ms' as the previous implementation did?

alecharp:
I don't feel that we should change this. @rsandell do you think we should close this PR?

PR #149, opened 2020-06-04
This tests are failing when running on the PCT

amuniz:
1. `FolderTest.shouldHaveHealthMetricConfiguredGloballyOnCreation`: is checking [something](https://github.com/jenkinsci/cloudbees-folder-plugin/blob/2b9443b5c5d72030a92f59e8843c8948022830b5/src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java#L420) which is not necessarily true (it depends on [this condition](https://github.com/jenkinsci/cloudbees-folder-plugin/blob/ae00f49fc0c6183983d67c799772fd8b9adc0840/src/main/java/com/cloudbees/hudson/plugins/folder/config/AbstractFolderConfiguration.java#L50)). So I suspect there is a `FolderHealthMetricDescriptor` in the classpath (from some other plugin) which is returning `null` on `FolderHealthMetricDescriptor#createDefault` (which is fine).

2. `AbstractFolderConfigurationTest#shouldBeAbleToRemoveHealthMetricConfiguredGlobally`: same reason.

* ...

### Submitter checklist

@fcojfernandez

### DIFF
--- a/src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
+++ b/src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
@@ -417,7 +417,7 @@ public class FolderTest {
     @Issue("JENKINS-58282")
     @Test public void shouldHaveHealthMetricConfiguredGloballyOnCreation() throws Exception {
         assertThat("by default, global configuration should have all folder health metrics",
-                AbstractFolderConfiguration.get().getHealthMetrics(), hasSize(FolderHealthMetricDescriptor.all().size()));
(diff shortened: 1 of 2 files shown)


amuniz:
The actual failure is happening on PCT runs (https://github.com/jenkinsci/bom/pull/225 for example)

timja:
@fcojfernandez would you be able to release these changes? or should we ignore the tests in bom for now?

fcojfernandez:
I plan to do it tomorrow or next Monday morning (CET) at the latest.

fcojfernandez:
https://github.com/jenkinsci/cloudbees-folder-plugin/releases/tag/cloudbees-folder-6.13 is out

jetersen:
Thanks 🙇

PR #150, opened 2020-06-09
[JENKINS-62632] Process all children before raising exceptions

varyvol:
See [JENKINS-62632](https://issues.jenkins-ci.org/browse/JENKINS-62632).

* Entry 1: Process and try to delete all children before actually raising the exceptions for those who could not be deleted.

### Submitter checklist

@fcojfernandez

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
@@ -67,6 +67,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -282,16 +283,26 @@ public abstract class ComputedFolder<I extends TopLevelItem> extends AbstractFol
                             new Object[]{getFullName(), orphaned.keySet()});
                 }
                 Collection<I> forRemoval = orphanedItems(orphaned.values(), listener);
+
+                List<IOException> deletingExceptions = new LinkedList<>();
                 for (I existing : orphaned.values()) {
                     if (forRemoval.contains(existing)) {
                         if (LOGGER.isLoggable(Level.FINE)) {
                             LOGGER.log(Level.FINE, "{0}: deleting {1}", new Object[]{getFullName(), existing});
                         }
-                        existing.delete();
-                        // super.onDeleted handles removal from items
+                        try {
+                            existing.delete();
+                            // super.onDeleted handles removal from items
+                        } catch (IOException x) { // InterruptedException is still propagated
+                            deletingExceptions.add(x);
+                        }
                     } else {
                         applyDisabled(existing, true);
                     }
+                }
+
+                if (!deletingExceptions.isEmpty()) {
+                    throw new DeletingChildrenException(deletingExceptions);
                 }
             }
         }
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/computed/DeletingChildrenException.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/DeletingChildrenException.java
@@ -14,7 +14,7 @@ public class DeletingChildrenException extends RuntimeException {
 
     @Override
     public String getMessage() {
-        return "Impossible to delete some children when computing the folder: \n" +
-                String.join("\n", exceptions.stream().map(e -> e.getMessage()).collect(Collectors.toList()));
+        return "Impossible to delete some children when computing the folder: \n\t" +
+                String.join("\n\t", exceptions.stream().map(e -> e.getMessage()).collect(Collectors.toList()));
     }
 }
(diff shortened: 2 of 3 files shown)


fcojfernandez:
Closing and re-opening to launch again the CI

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DeletingChildrenException.java
```diff
@@ -0,0 +1,20 @@
+package com.cloudbees.hudson.plugins.folder.computed;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class DeletingChildrenException extends RuntimeException {
+
+    private List<IOException> exceptions;
```
MRamonLeon, line 9: would you consider to store just the message?

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
```diff
@@ -282,17 +283,27 @@ final void updateChildren(final TaskListener listener) throws IOException, Inter
...
                     if (forRemoval.contains(existing)) {
                         if (LOGGER.isLoggable(Level.FINE)) {
                             LOGGER.log(Level.FINE, "{0}: deleting {1}", new Object[]{getFullName(), existing});
                         }
-                        existing.delete();
-                        // super.onDeleted handles removal from items
+                        try {
+                            existing.delete();
+                            // super.onDeleted handles removal from items
+                        } catch (IOException x) { // InterruptedException is still propagated
```
MRamonLeon, line 296: why do we want to stop with something different than IOException?

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DeletingChildrenException.java
varyvol: I stored the exceptions just in case they could be useful in the future.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
varyvol: The other possibility is InterruptedException and it seems weird to continue if that's the case.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DeletingChildrenException.java
alecharp: The message wouldn't be enough to understand which part of the code triggered the exception.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DeletingChildrenException.java
```diff
@@ -0,0 +1,20 @@
...
+
+    public DeletingChildrenException(List<IOException> exceptions) {
+        this.exceptions = exceptions;
+    }
+
+    @Override
+    public String getMessage() {
+        return "Impossible to delete some children when computing the folder: \n" +
+                String.join("\n", exceptions.stream().map(e -> e.getMessage()).collect(Collectors.toList()));
```
alecharp, line 18: Do you think adding some indentation in the message could visually help to understand the end of the list?
```suggestion
                String.join("\n\t", exceptions.stream().map(e -> e.getMessage()).collect(Collectors.toList()));
```

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DeletingChildrenException.java
```diff
@@ -0,0 +1,20 @@
+package com.cloudbees.hudson.plugins.folder.computed;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class DeletingChildrenException extends RuntimeException {
```
rsandell, line 7: This should be an `IOException` with the cause set to the first exception in the list.
That way it would be much easier finding the reason for the error as the stack trace in the log would show the first error as well, and the likelihood of problems deleting are the same is very high.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DeletingChildrenException.java
rsandell: If you have hundreds of items failing to be deleted this will just clog up the logs making it hard to read and possibly filling up the buffer. You should at least make the list of messages distinct (though they might not be since they might contain the path of the file)

varyvol: @rsandell so do you think the message would be better removed?

varyvol:
@rsandell can you please specify what do you mean with downstream testing?

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DeletingChildrenException.java
rsandell: Yes, If you add the first in the list as the cause then we can still see that in the stacktrace.
Or if you'd like take for example only the first three distinct messages.

rsandell:
@varyvol Testing that it works well in GitHub multibranch for example, just some manual testing and/or running the unit tests with the new version to confirm nothing is broken in what that plugin expects.

varyvol:
@rsandell @fcojfernandez I've changed to IOException together with using the first cause plus limiting the number of messages to 3. That way a flooding is avoided, several items that could not be deleted are displayed together with the cause for the first one as well. 

Screenshot 2020-06-10 at 11 24 42

I've also executed the tests for branch-api with this version and everything works fine.

fcojfernandez:
@rsandell I think this PR is ready. I've re-request your review since you requested changes, so please review it again

PR #153, opened 2020-10-02
[JENKINS-63836] Not add heath metrics by default in the Global Configuration

fbelzunc:
See [JENKINS-63836](https://issues.jenkins-ci.org/browse/JENKINS-63836).

* [JENKINS-63836](https://issues.jenkins-ci.org/browse/JENKINS-63836) : Not add heath metrics by default in the Global Configuration

### Submitter checklist

@fcojfernandez

### DIFF
--- a/src/test/java/com/cloudbees/hudson/plugins/folder/FolderSystemPropertyTest.java
+++ b/src/test/java/com/cloudbees/hudson/plugins/folder/FolderSystemPropertyTest.java
@@ -45,7 +45,7 @@ public class FolderSystemPropertyTest {
     public JenkinsRule j = new JenkinsRule();
 
     @BeforeClass
-    public static void enableHealhMetrics() {
+    public static void enableHealthMetrics() {
         System.setProperty(AbstractFolderConfiguration.class.getName() + ".ADD_HEALTH_METRICS", "true");
     }
 
--- a/src/test/java/com/cloudbees/hudson/plugins/folder/FolderSystemPropertyTest.java
+++ b/src/test/java/com/cloudbees/hudson/plugins/folder/FolderSystemPropertyTest.java
@@ -51,7 +51,7 @@ public class FolderSystemPropertyTest {
 
     @AfterClass
     public static void disableHealthMetrics() {
-        System.setProperty(AbstractFolderConfiguration.class.getName() + ".ADD_HEALTH_METRICS", null);
+        System.clearProperty(AbstractFolderConfiguration.class.getName() + ".ADD_HEALTH_METRICS");
     }
 
     @Issue("JENKINS-63836")
--- a/src/test/java/com/cloudbees/hudson/plugins/folder/FolderSystemPropertyTest.java
+++ b/src/test/java/com/cloudbees/hudson/plugins/folder/FolderSystemPropertyTest.java
@@ -44,14 +44,22 @@ public class FolderSystemPropertyTest {
     @Rule
     public JenkinsRule j = new JenkinsRule();
 
+    private static String HEALTH_METRIC_PROPERTY;
+
     @BeforeClass
     public static void enableHealthMetrics() {
+        HEALTH_METRIC_PROPERTY = System.getProperty(AbstractFolderConfiguration.class.getName() + ".ADD_HEALTH_METRICS");
         System.setProperty(AbstractFolderConfiguration.class.getName() + ".ADD_HEALTH_METRICS", "true");
     }
 
     @AfterClass
     public static void disableHealthMetrics() {
-        System.clearProperty(AbstractFolderConfiguration.class.getName() + ".ADD_HEALTH_METRICS");
+        // Put back the previous value before the test was executed
+        if (HEALTH_METRIC_PROPERTY != null) {
+            System.setProperty(AbstractFolderConfiguration.class.getName() + ".ADD_HEALTH_METRICS", HEALTH_METRIC_PROPERTY);
+        } else {
+            System.clearProperty(AbstractFolderConfiguration.class.getName() + ".ADD_HEALTH_METRICS");
+        }
     }
 
     @Issue("JENKINS-63836")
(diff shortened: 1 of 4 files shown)


dariver:
👍  not adding the health metrics by default. 
But instead of removing the old code, maybe it's better to keep the old behavior by setting up a system property.

fbelzunc:
@dariver Very good idea. I just added the system property to keep the old behavior.

@fcojfernandez I think this is ready to be reviewed.

src/test/java/com/cloudbees/hudson/plugins/folder/config/AbstractFolderConfigurationTest.java
```diff
@@ -43,10 +47,21 @@ public void testInitialization() {
...
+
     @Issue("JENKINS-60393")
     @Test
     public void shouldBeAbleToRemoveHealthMetricConfiguredGlobally() throws Exception {
-        assertThat("by default, global configuration should have all folder health metrics",
+        List<FolderHealthMetric> healthMetrics = new ArrayList<>(1);
+        healthMetrics.add(new WorstChildHealthMetric(true));
+        AbstractFolderConfiguration.get().setHealthMetrics(healthMetrics);
+
+        assertThat("by default, global configuration should not have any folder health metrics",
```
fcojfernandez, line 64: Remove this assertion since **now** it doesn't make sense and it's already checked in `FolderTest`

src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
```diff
@@ -425,19 +425,19 @@ private void copyViaHttp(Folder f, JenkinsRule.WebClient wc, String fromName, St
 
     @Issue("JENKINS-58282")
     @Test public void shouldHaveHealthMetricConfiguredGloballyOnCreation() throws Exception {
-        assertThat("by default, global configuration should have all folder health metrics",
-                AbstractFolderConfiguration.get().getHealthMetrics(), hasSize((int) FolderHealthMetricDescriptor.all().stream().filter(d -> d.createDefault() != null).count()));
+        assertThat("by default, global configuration should not have any health metrics",
+                AbstractFolderConfiguration.get().getHealthMetrics(), hasSize(0));
```
fcojfernandez, line 429: Because now the global configuration shouldn't have any health metrics by default, rename the test from `shouldHaveHealthMetricConfiguredGloballyOnCreation` to `shouldNotHaveHealthMetricConfiguredGloballyOnCreation`.

fcojfernandez (approved):

src/test/java/com/cloudbees/hudson/plugins/folder/FolderTest.java
```diff
@@ -424,20 +424,20 @@ private void copyViaHttp(Folder f, JenkinsRule.WebClient wc, String fromName, St
     }
 
     @Issue("JENKINS-58282")
-    @Test public void shouldHaveHealthMetricConfiguredGloballyOnCreation() throws Exception {
-        assertThat("by default, global configuration should have all folder health metrics",
-                AbstractFolderConfiguration.get().getHealthMetrics(), hasSize((int) FolderHealthMetricDescriptor.all().stream().filter(d -> d.createDefault() != null).count()));
+    @Test public void shouldNotHaveHealthMetricConfiguredGloballyOnCreation() throws Exception {
```
varyvol, line 427: Is it possible to create a test that validates the metric appears when the property is set? (basically the previous test plus setting the property to true).

fcojfernandez: @fbelzunc I think @varyvol has made a great point!

fbelzunc: @varyvol @fcojfernandez Sure, good idea! Do you want me to set the property with `System.setProperty` or there is something better I am not aware of?

fcojfernandez: `System.setProperty` is OK. Just remember to clear the value after the test has finished.

src/test/java/com/cloudbees/hudson/plugins/folder/FolderSystemPropertyTest.java
```diff
@@ -0,0 +1,76 @@
...
+        DescribableList<FolderHealthMetric, FolderHealthMetricDescriptor> healthMetrics = folder.getHealthMetrics();
+        assertThat("a new created folder should have all the folder health metrics configured globally",
+                healthMetrics.toList(), containsInAnyOrder(AbstractFolderConfiguration.get().getHealthMetrics().toArray()));
+
+        AbstractFolderConfiguration.get().setHealthMetrics(null);
+        folder = j.jenkins.createProject(Folder.class, "myFolder2");
+        healthMetrics = folder.getHealthMetrics();
+        assertThat("a new created folder should have all the folder health metrics configured globally",
+                healthMetrics, iterableWithSize(0));
+        System.clearProperty(AbstractFolderConfiguration.class.getName() + ".ADD_HEALTH_METRICS");
```
varyvol, line 73: This is not necessary anymore, right?

src/test/java/com/cloudbees/hudson/plugins/folder/FolderSystemPropertyTest.java
```diff
@@ -0,0 +1,76 @@
...
+    public JenkinsRule j = new JenkinsRule();
+
+    @BeforeClass
+    public static void enableHealthMetrics() {
+        System.setProperty(AbstractFolderConfiguration.class.getName() + ".ADD_HEALTH_METRICS", "true");
+    }
+
+    @AfterClass
+    public static void disableHealthMetrics() {
+        System.clearProperty(AbstractFolderConfiguration.class.getName() + ".ADD_HEALTH_METRICS");
```
varyvol, line 54: I know it's kind of picky, but it would be better if this would restore the original value (imagine you are running this on the system where Jenkins is running).

fbelzunc: @varyvol I am not sure what you are looking for. I think the test is already doing what you are asking for.

fbelzunc: @varyvol I think I finally understood your feedback and I just made a commit with the corresponded change. I think the PR should be fine now. Please, review when you have some spare time :-)

varyvol (approved):

PR #165, opened 2020-11-01
JENKINS-64069 Fix plugin / branch-api on 2.264

timja:
See [JENKINS-64069](https://issues.jenkins-ci.org/browse/JENKINS-64069).

As well as branch-api fixs this plugin needs updates otherwise it breaks branch-api

### Proposed changelog entries

* Entry 1: JENKINS-64069, Add compatibility with Jenkins 2.264
* ...

### Submitter checklist

- [ ] JIRA issue is well described
- [ ] Changelog entry appropriate for the audience affected by the change (users or developer, depending on the change).
      * Use the `Internal: ` prefix if the change has no user-visible impact (API, test frameworks, etc.)
- [ ] Appropriate autotests or explanation to why this change has no tests

@fcojfernandez

PR #169, opened 2020-11-10
Refresh POM and dependencies

jglick:
See https://github.com/jenkinsci/cloudbees-folder-plugin/pull/165#issuecomment-724716790. Supersedes #166, #167, and #168.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,1 @@
-buildPlugin(configurations: buildPlugin.recommendedConfigurations())
+buildPlugin(useAci: true)
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>4.2</version>
+    <version>4.13</version>
     <relativePath />
   </parent>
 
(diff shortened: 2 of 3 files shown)


Jenkinsfile
```diff
@@ -1 +1 @@
-buildPlugin(configurations: buildPlugin.recommendedConfigurations())
+buildPlugin(useAci: true)
```
fcojfernandez, line 1: Why are you removing the `buildPlugin.recommendedConfigurations()`? I thought that was the correct way to execute the `buildPlugin` function

jglick: https://github.com/jenkins-infra/pipeline-library/pull/145 and in particular it is impossible for someone without write access to the repo to accurately test changes to `jenkins.version` so long as the `Jenkinsfile` refers to other Jenkins versions.

timja (approved):

Jenkinsfile
fcojfernandez: Thanks for the explanation!

fcojfernandez (approved):

pom.xml
```diff
@@ -21,10 +21,10 @@
   <properties>
     <revision>6.15</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.176.1</jenkins.version>
+    <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
+    <jenkins.version>2.204.6</jenkins.version>
```
jglick, line 25: see https://github.com/jenkinsci/snakeyaml-api-plugin/pull/2#discussion_r473643918

PR #174, opened 2021-01-09
Use Java 7 versions of GZIPInputStream and GZIPOutputStream

basil:
See jenkinsci/docker-build-step-plugin#75. [JZlib](https://github.com/ymnk/jzlib) has not been updated since 2013. Since then, [`GZIPInputStream`](https://docs.oracle.com/javase/8/docs/api/java/util/zip/GZIPInputStream.html) and [`GZIPOutputStream`](https://docs.oracle.com/javase/8/docs/api/java/util/zip/GZIPOutputStream.html) have been added to the Java 7 API, so it seems more prudent to use those implementations.

CC @fcojfernandez


basil:
Oh wow, we actually switched from the Java API _to_ JZlib in [JENKINS-14362](https://issues.jenkins.io/browse/JENKINS-14362) for performance reasons. Well never mind, then!

PR #185, opened 2021-07-19
[JENKINS-66168] Fix the Disable/Enable button for non primary views

Dohbedoh:
See [JENKINS-66168](https://issues.jenkins.io/browse/JENKINS-66168)

* Fix "Enable|Disable Multibranch Pipeline" button from non-primary views of Folders.

### Submitter checklist


Dohbedoh:
Wanted to demonstrate that the test fails without the fix but I think that the second build maybe interrupted the first one "Superseded by Plugins/cloudbees-folder-plugin/PR-185#2" :(

Dohbedoh:
ping @fcojfernandez for review

bitwiseman (approved):

PR #196, opened 2021-10-26
[JENKINS-34502] Provide health metric for a named child of the folder

strangelookingnerd:
See [JENKINS-34502](https://issues.jenkins-ci.org/browse/JENKINS-34502).

* Entry 1: JENKINS-34502, Provide health metric for a named child of the folder

### Submitter checklist

### DIFF
--- a/src/test/java/com/cloudbees/hudson/plugins/folder/health/NamedChildHealthMetricTest.java
+++ b/src/test/java/com/cloudbees/hudson/plugins/folder/health/NamedChildHealthMetricTest.java
@@ -25,7 +25,7 @@
 package com.cloudbees.hudson.plugins.folder.health;
 
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.collection.IsCollectionWithSize.hasSize
+import static org.hamcrest.collection.IsCollectionWithSize.hasSize;
 
 import java.util.List;
 


strangelookingnerd:
Can I get a feedback please?

@fcojfernandez @bitwiseman @alecharp

jtnord (changes requested):
at least the Jira this references has been solved for a long time - thus I am wondering if this is even required.

For multibranch projects, there is the ability to only report the health of the main branch for a repository.  thus this PR only grants the extra ability to report the health of a specific named child .   As this seems to be very edge case I would prefer that this would not be in this plugin but a separate plugin if it is deemed that this is required and was not due to missing the configuration option that can do this today for multi-branch projects.

strangelookingnerd:
> For multibranch projects, there is the ability to only report the health of the main branch for a repository. thus this PR only grants the extra ability to report the health of a specific named child . As this seems to be very edge case I would prefer that this would not be in this plugin but a separate plugin if it is deemed that this is required and was not due to missing the configuration option that can do this today for multi-branch projects.

This is mostly true, however not all SCM (and respective Jenkins SCM Providers) do have a concept for "primary branches".
Further there is still the possibility that one does not not have a multi-branch project but a "simple" folder with "loose" build jobs or even nested structures of folders. I myself am missing that feature, can't really tell if others feel the same about it tho ;)

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -894,68 +894,62 @@ public void invalidateBuildHealthReports() {
...
+        if (type.isWithChildren()) {
+            if (type.isRecursive()) {
+                Stack<Iterable<? extends Item>> stack = new Stack<Iterable<? extends Item>>();
+                stack.push(getItems());
+                if (type.isTopLevelItems()) {
+                    while (!stack.isEmpty()) {
+                        for (Item item : stack.pop()) {
+                            if (item instanceof TopLevelItem) {
                                 reporter.observe(item);
+                                if (item instanceof Folder) {
```
jtnord, line 929: Not every container is a `Folder`  why checking here for `Folder` rather than `AbstractFolder`?

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -894,68 +894,62 @@ public void invalidateBuildHealthReports() {
...
-                    for (Item item : stack.pop()) {
-                        if (item instanceof TopLevelItem) {
-                            for (FolderHealthMetric.Reporter reporter : reporters) {
+
+        Collections.sort(reports);
+        healthReports = reports; // idempotent write
+        return reports;
+    }
+
+    private void observeMetric(FolderHealthMetric.Type type, FolderHealthMetric.Reporter reporter) {
```
jtnord, line 919: this seems to change how the health is computed.

Before this change IIUC if on metric ype was recursive or one properties health metric type was recursive, then all the metrics will be calculated recursivley, same as if one metric was `topLevelOnly` this would apply to everything.

I am not convinced that the existing behaviour seems correct, but it seems like a change like that that should be decoupled from this PR?

jtnord (commented):
couple of questions on the code, one which looks like a change of behaviour.

Would it be possible to add a unit test to show this new metric (I appreciate that in general folder health is not covered well / at all) but demonstrating this for a folder that does does not contain the named child but does contain others, does contain the named child, and contains the named child at multiple depths, should be relatively simple and prevent any regressions.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
strangelookingnerd: Not sure to be honest. The check was for {{Folder}} before my changes - but I agree that it should be changed to {{AbstractFolder}}. I will test it and report back.

jtnord: OK, in which case, may be better to leave it as it.
Either it is an existing bug that we should fix separately, or there was a reason for it. Having as dedicated change as possible would help to identify any causes of regressions (and when we can spread that over multiple releases it makes it even better).

jtnord: OK, in which case, may be better to leave it as it.
Either it is an existing bug that we should fix separately, or there was a reason for it. Having as dedicated change as possible would help to identify any causes of regressions (and when we can spread that over multiple releases it makes it even better).

strangelookingnerd:
I added a couple tests as requested.

jtnord (approved):

strangelookingnerd:
@jtnord Thanks for your suggestions, really appreciated. Haven't worked with hamcrest, I'm more of a JUnit / JAssert kinda guy ;)

src/test/java/com/cloudbees/hudson/plugins/folder/health/NamedChildHealthMetricTest.java
```diff
@@ -0,0 +1,76 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.cloudbees.hudson.plugins.folder.health;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.collection.IsCollectionWithSize.hasSize
```
jtnord, line 28:
```suggestion
import static org.hamcrest.collection.IsCollectionWithSize.hasSize;
```

strangelookingnerd:
May I ask why this has been closed? I was under the impressions that this is a useful enhancement :(

jglick:
Not sure what caused it to be closed; GitHub does not say. Did you accidentally delete your fork branch or something?

strangelookingnerd:
You are indeed correct. I must have deleted it by accident. I'll recover it and re-create the PR.

PR #197, opened 2021-11-07
[JENKINS-60677] Option to cancel builds when orphaning items

reda-alaoui:
See [JENKINS-60677](https://issues.jenkins-ci.org/browse/JENKINS-60677).

* Entry 1: JENKINS-60677, Add an option allowing to cancel builds of orphaned items.

### Submitter checklist


jtnord (changes requested):
I am concerned about waiting indefinatly for something that may never occur (a build finishing correctly when asked) along with handling things still in the queue and pipelines.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
```diff
@@ -71,33 +74,59 @@
...
         this.numToKeep = pruneDeadBranches ? fromString(numToKeepStr) : -1;
+        this.abortBuilds = abortBuilds;
+    }
+
+    /**
+     * @deprecated Use {@link #DefaultOrphanedItemStrategy(boolean, int, int, boolean)} instead
+     */
+    @Deprecated
+    public DefaultOrphanedItemStrategy(boolean pruneDeadBranches, int daysToKeep, int numToKeep) {
+        this(pruneDeadBranches, daysToKeep, numToKeep, false);
```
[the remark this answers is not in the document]
jglick: https://www.jenkins.io/doc/developer/plugin-development/pipeline-integration/#constructor-vs-setters

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
```diff
@@ -230,6 +269,27 @@ private static boolean disabled(TopLevelItem item) {
 
     @Override
     public <I extends TopLevelItem> Collection<I> orphanedItems(ComputedFolder<I> owner, Collection<I> orphaned, TaskListener listener) throws IOException, InterruptedException {
+        if (abortBuilds) {
+            Queue<Run<?, ?>> abortedBuilds = new LinkedList<>();
+            for (I item: orphaned) {
+                for (Job<?, ?> job: item.getAllJobs()) {
+                    for (Run<?, ?> build: job.getBuilds()) {
```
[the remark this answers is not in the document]
jglick: Indeed you should not do this. Cut off the search when encountering a completed build. (Not totally correct, since it is possible for there to be a running build older than the last completed build, but good enough as a heuristic.) The iterator is lazy so you can keep the loop as is, so long as there is a break condition.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
```diff
@@ -230,6 +269,27 @@ private static boolean disabled(TopLevelItem item) {
...
+                        Executor executor = build.getExecutor();
+                        if (executor == null) {
```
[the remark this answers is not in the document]
jglick: You would need a separate code path looking for `Queue.Item`’s associated with the job and `cancel`’ing them.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
```diff
@@ -230,6 +269,27 @@ private static boolean disabled(TopLevelItem item) {
...
+                            continue;
+                        }
+                        executor.interrupt(Result.ABORTED);
```
jglick, line 281: Best to define a new `CauseOfInterruption`.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
```diff
@@ -230,6 +269,27 @@ private static boolean disabled(TopLevelItem item) {
 
     @Override
     public <I extends TopLevelItem> Collection<I> orphanedItems(ComputedFolder<I> owner, Collection<I> orphaned, TaskListener listener) throws IOException, InterruptedException {
+        if (abortBuilds) {
+            Queue<Run<?, ?>> abortedBuilds = new LinkedList<>();
+            for (I item: orphaned) {
+                for (Job<?, ?> job: item.getAllJobs()) {
+                    for (Run<?, ?> build: job.getBuilds()) {
+                        Executor executor = build.getExecutor();
```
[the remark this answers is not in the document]
jglick: I believe this will work for Pipeline builds, but indeed it would be nice to have test coverage proving it.

Would I think need a `test`-scoped dep on `workflow-multibranch` and the `tests` classifier of `scm-api` to make a multibranch project governed by a mock branch source that you can manipulate.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
```diff
@@ -230,6 +269,27 @@ private static boolean disabled(TopLevelItem item) {
...
+                        abortedBuilds.add(build);
+                    }
+                }
+            }
+            Run<?, ?> abortedBuild;
+            while ((abortedBuild = abortedBuilds.poll()) != null) {
+                while (abortedBuild.isLogUpdated()) {
+                    Thread.sleep(100L);
+                }
+            }
```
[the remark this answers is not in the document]
jglick: I guess we are waiting because https://github.com/jenkinsci/cloudbees-folder-plugin/blob/53bf95c744c8a0ff468ef6d2b699796d2186075e/src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java#L318 relies on it? Definitely there should be a time limit waiting for an interrupt to be processed. It is not generally a big deal if we have to skip one of the orphaned projects because it has a build still running; it should just be processed the next time branch indexing runs.

src/main/resources/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy/config.jelly
```diff
@@ -23,6 +23,10 @@
  -->
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
+  <f:entry title="${%Abort builds}" field="abortBuilds"
+      description="Abort all pending or ongoing builds">
+    <f:checkbox/>
```
jglick, line 28:
```suggestion
  <f:entry field="abortBuilds">
    <f:checkbox title="${%Abort builds}"/>
```

conventionally, but make sure to add a `help-abortBuilds.html` with details.

jglick (commented):
Looking promising.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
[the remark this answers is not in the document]
reda-alaoui: Ceiled with a max wait of 1 minute for all aborted builds.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
```diff
@@ -273,19 +259,35 @@ private static boolean disabled(TopLevelItem item) {
             Queue<Run<?, ?>> abortedBuilds = new LinkedList<>();
             for (I item: orphaned) {
                 for (Job<?, ?> job: item.getAllJobs()) {
+                    if (job instanceof hudson.model.Queue.Task) {
+                        hudson.model.Queue jenkinsQueue = Jenkins.get().getQueue();
+                        hudson.model.Queue.Task task = (hudson.model.Queue.Task) job;
+                        for (hudson.model.Queue.Item pendingBuild: jenkinsQueue.getItems(task)) {
+                            jenkinsQueue.cancel(pendingBuild);
```
jglick, line 266: FYI possibly already covered by https://github.com/jenkinsci/jenkins/blob/1a42044347502c59cd9c39d1be67f654df02908a/core/src/main/java/jenkins/model/ParameterizedJobMixIn.java#L486

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
```diff
@@ -273,19 +259,35 @@ private static boolean disabled(TopLevelItem item) {
...
+                            // the last completed build, but good enough as a heuristic.
+                            break;
                         }
-                        executor.interrupt(Result.ABORTED);
+                        executor.interrupt(Result.ABORTED, new OrphanedParent(item));
                         abortedBuilds.add(build);
                     }
                 }
             }
+            long waitForAbortedBuildsStartTimeMillis = System.currentTimeMillis();
```
jglick, line 283: use `System.nanoTime` for comparisons

src/main/java/com/cloudbees/hudson/plugins/folder/computed/OrphanedParent.java
```diff
@@ -0,0 +1,46 @@
...
+ * Indicates that an execution was aborted because of the orphaning of one of its parents.
+ *
+ * @author Réda Housni Alaoui
+ */
+public class OrphanedParent extends CauseOfInterruption {
+
+	private final String orphanedParentFullName;
+
+	public OrphanedParent(Item orphanedParent) {
+		this.orphanedParentFullName = orphanedParent.getFullName();
```
jglick, line 39:
```suggestion
		this.orphanedParentFullName = orphanedParent.getFullDisplayName();
```

(if indeed used only for display)

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
```diff
@@ -91,8 +101,8 @@ public DefaultOrphanedItemStrategy(boolean pruneDeadBranches, @CheckForNull Stri
      * Programmatic constructor.
      *
      * @param pruneDeadBranches remove dead branches.
-     * @param daysToKeep     how old a branch must be to remove.
-     * @param numToKeep      how many branches to keep.
+     * @param daysToKeep        how old a branch must be to remove.
+     * @param numToKeep         how many branches to keep.
```
jglick, line 105:
```suggestion
     * @param daysToKeep     how old a branch must be to remove.
     * @param numToKeep      how many branches to keep.
```

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
```diff
@@ -230,6 +255,43 @@ private static boolean disabled(TopLevelItem item) {
...
+                        }
+                        executor.interrupt(Result.ABORTED, new OrphanedParent(item));
+                        abortedBuilds.add(build);
+                    }
+                }
+            }
+            long waitForAbortedBuildsStartTimeMillis = System.currentTimeMillis();
+            long maxWaitMillis = 60_000;
+            Run<?, ?> abortedBuild;
+            ABORTED_BUILDS: while ((abortedBuild = abortedBuilds.poll()) != null) {
```
jglick, line 286: Why the queue rather than just iterating the list with a simple `for`-loop?

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
```diff
@@ -230,6 +255,43 @@ private static boolean disabled(TopLevelItem item) {
...
+                while (abortedBuild.isLogUpdated()) {
+                    if ((System.currentTimeMillis() - waitForAbortedBuildsStartTimeMillis) > maxWaitMillis) {
+                        break ABORTED_BUILDS;
+                    }
+                    Thread.sleep(100L);
```
jglick, line 291: By now I have already forgotten why we need to wait for the aborted builds to complete at all…so it would be a good idea to add a comment explaining the need for the benefit of future generations.

src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
```diff
@@ -265,6 +272,99 @@ public void runningBuild() throws Exception {
...
+    @Test
+    public void runningWorflowJobBuildWithAbortBuildsOption() throws Exception {
+        SampleComputedFolderWithWorkflowJobAsChildren folder = r.jenkins.createProject(SampleComputedFolderWithWorkflowJobAsChildren.class, "d");
+        DefaultOrphanedItemStrategy strategy = new DefaultOrphanedItemStrategy(true, -1, -1);
+        strategy.setAbortBuilds(true);
+        folder.setOrphanedItemStrategy(strategy);
+        folder.kids.add("kid");
+        folder.recompute(Result.SUCCESS);
+        folder.assertItemNames(1, "kid");
+        folder.getItem("kid").setDefinition(new CpsFlowDefinition("node { sleep(Long.MAX_VALUE) }", true));
```
jglick, line 322: Or just

```suggestion
        folder.getItem("kid").setDefinition(new CpsFlowDefinition("sleep(Long.MAX_VALUE)", true));
```

should suffice I think.

And does this really work at all? You do not have a test dep on `workflow-basic-steps` nor `workflow-durable-task-step`.

If you add a test dep on the `tests` classified of `workflow-support` then you get the `semaphore` test step which is handy for things like this: can have the test wait for the build to get a well-defined point and pause before proceeding.

src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
```diff
@@ -740,6 +840,68 @@ public TopLevelItem newInstance(ItemGroup parent, String name) {
 
     }
 
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    public static class SampleComputedFolderWithWorkflowJobAsChildren extends ComputedFolder<WorkflowJob> {
```
jglick, line 844: What is the point of the `workflow-multibranch` test dep if you are not actually creating a `MultiBranchProject`?

jglick (commented):
Code changes look right to me. Tests could use some work. Sorry for the long delay in reviews, I am a bit overwhelmed, but I just remembered this PR and that it was important so I will try to help more promptly.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
reda-alaoui: I was going to use this at first, but then I noticed `Queue#cancel(Task)` only cancel the first item of the list:
```java
    /**
     * Cancels the item in the queue. If the item is scheduled more than once, cancels the first occurrence.
     *
     * @return true if the project was indeed in the queue and was removed.
     *         false if this was no-op.
     */
    public boolean cancel(Task p) {
        lock.lock();
        try { try {
            LOGGER.log(Level.FINE, "Cancelling {0}", p);
            for (WaitingItem item : waitingList) {
                if (item.task.equals(p)) {
                    return item.cancel(this);
                }
            }
            // use bitwise-OR to make sure that both branches get evaluated all the time
            return blockedProjects.cancel(p) != null | buildables.cancel(p) != null;
        } finally { updateSnapshot(); } } finally {
            lock.unlock();
        }
    }
```
Not being acquainted to the Jenkins model, I chose the safety by cancelling the list of items.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
reda-alaoui: Mostly for conceptual reasons. Also, polling on a queue frees memory faster.

src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
reda-alaoui: You are correct, the build was doomed to fail but couldn't because abort was too fast.
I fixed that with the `semaphore` step.

src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
reda-alaoui: None indeed. Dependencies cleaned.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
jglick: Probably safest to redundantly cancel all items as you do here. In practice it would rarely matter since queue items would be coalesced; they would not be when using parameterized builds, but of course branch projects are much less likely to be parameterized.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
jglick: What is the concept here? And why would we care about a few bytes of a `LinkedList`? Seems like needless complexity.

src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
```diff
@@ -319,7 +319,7 @@ public void runningWorflowJobBuildWithAbortBuildsOption() throws Exception {
         folder.kids.add("kid");
         folder.recompute(Result.SUCCESS);
         folder.assertItemNames(1, "kid");
-        folder.getItem("kid").setDefinition(new CpsFlowDefinition("node { sleep(Long.MAX_VALUE) }", true));
+        folder.getItem("kid").setDefinition(new CpsFlowDefinition("semaphore('foo')", true));
```
jglick, line 322: This pairs with `SemaphoreStep.waitForStart`.

src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
jglick, line 844: Note: not as realistic as using `workflow-multibranch`, but simpler to set up and probably good enough for this purpose.

src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
reda-alaoui: Done

src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
```diff
@@ -265,6 +273,101 @@ public void runningBuild() throws Exception {
...
+        folder.recompute(Result.SUCCESS);
+        folder.assertItemNames(4, "A");
+        aBuild1.keepLog(false);
+        folder.recompute(Result.SUCCESS);
+        folder.assertItemNames(5);
+    }
+
+    @Issue("JENKINS-60677")
+    @Test
+    public void runningWorflowJobBuildWithAbortBuildsOption() throws Exception {
```
jglick, line 315:
```suggestion
    public void runningWorkflowJobBuildWithAbortBuildsOption() throws Exception {
```

jglick: or
```suggestion
    public void runningPipelineBuildWithAbortBuildsOption() throws Exception {
```

src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
```diff
@@ -265,6 +273,101 @@ public void runningBuild() throws Exception {
...
+        SampleComputedFolderWithWorkflowJobAsChildren folder = r.jenkins.createProject(SampleComputedFolderWithWorkflowJobAsChildren.class, "d");
+        DefaultOrphanedItemStrategy strategy = new DefaultOrphanedItemStrategy(true, -1, -1);
+        strategy.setAbortBuilds(true);
+        folder.setOrphanedItemStrategy(strategy);
+        folder.kids.add("kid");
+        folder.recompute(Result.SUCCESS);
+        folder.assertItemNames(1, "kid");
+        folder.getItem("kid").setDefinition(new CpsFlowDefinition("semaphore('wait')", true));
+        folder.getItem("kid").setConcurrentBuild(true);
```
jglick, line 324: I think this is the default for Pipeline anyway, so could
```suggestion
```

jglick (approved):
Nice! Will leave it to a plugin maintainer to merge. @jtnord who would that be?

reda-alaoui:
Ping :)

olamy (approved):

reda-alaoui:
Ping :pray:

reda-alaoui:
Merge conflict fixed.

reda-alaoui:
Looks like f6905c07c8e1d3ca204f193062fd1fd2f253cbf5 changed the `pom.xml` end of line from LF to CRLF in master. Was that wanted? Do I have to change the `pom.xml` line ending in this PR?

jglick:
> Was that wanted?

No. https://issues.apache.org/jira/browse/MRELEASE-899

> Do I have to change the `pom.xml` line ending to CRLF in this PR?

No, let @jtnord fix it up.

jtnord:
> > Was that wanted?
> 
> No. https://issues.apache.org/jira/browse/MRELEASE-899
> 
> > Do I have to change the `pom.xml` line ending to CRLF in this PR?
> 
> No, let @jtnord fix it up.

Will do, I started walking though these to merge & release, will continue tomorrow.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
```diff
@@ -230,6 +253,46 @@ private static boolean disabled(TopLevelItem item) {
 
     @Override
     public <I extends TopLevelItem> Collection<I> orphanedItems(ComputedFolder<I> owner, Collection<I> orphaned, TaskListener listener) throws IOException, InterruptedException {
+        if (abortBuilds) {
+            List<Run<?, ?>> abortedBuilds = new ArrayList<>();
+            for (I item: orphaned) {
+                for (Job<?, ?> job: item.getAllJobs()) {
```
jtnord, line 259: what happens if this was not an organisation folder but a regular multibranch?

IIUC owner would be the Multibranch folder, and orphaned would be a WorkflowJob, so orphaned.getAllJobs() would be `null` or empty.

does this only work for an organisation folder when an entire repository is removed?

or is this handled somewhere else?

possibly I am missing something

jtnord (commented):
@reda-alaoui could you take a look at https://github.com/jenkinsci/cloudbees-folder-plugin/pull/197/files#r809952123

this is on my list of things to do, I am just currently swamped

src/main/java/com/cloudbees/hudson/plugins/folder/computed/DefaultOrphanedItemStrategy.java
jglick: https://github.com/jenkinsci/jenkins/blob/1a807197ba0c574c6f9c79803fcd1d5dc14fbf43/core/src/main/java/hudson/model/Job.java#L506-L509 ?

jtnord (approved):

PR #199, opened 2021-11-10
Fixes JENKINS-56935

ericcitaire:
See [JENKINS-56935](https://issues.jenkins-ci.org/browse/JENKINS-56935).

* Fixes JENKINS-56935

### Submitter checklist


ericcitaire:
Can I have feedback on this please ?

ping @fcojfernandez @bitwiseman @alecharp

jtnord (changes requested):
this would be incorrect as the code is setting the internal value to `-1`  when the value is empty.  by allowing `0` the internal field would no longer be using `-1` and the quick deletion would no longer work.

If the behaviour is unclear I would suggest a change to the inline help and documentation rather than code

ericcitaire:
> this would be incorrect as the code is setting the internal value to `-1` when the value is empty. by allowing `0` the internal field would no longer be using `-1` and the quick deletion would no longer work.
> 
> If the behaviour is unclear I would suggest a change to the inline help and documentation rather than code

Yes, the behaviour is unclear.
From what I observe, my understanding is that an empty value or `-1` basically means *ignore this criteria*, while a `0` value means *don't keep anything*.

Also, this is just a matter of UX because the `0` value is actually accepted, despite the displayed error message.

PR #200, opened 2021-11-11
Use SVG icons where possible

zbynek:
Based on https://github.com/jenkinsci/jenkins/pull/5663, update some folder/view specific icons.
* icons referenced from core by class instead of filename -> will use SVGs with core 2.308 and later
* icons provided by this plugin available as SVG, PNGs kept in case some dependent plugin uses them

* Use SVG icons where possible
* ...

### Submitter checklist

- [n/a] JIRA issue is well described
- [n/a] Appropriate autotests or explanation to why this change has no tests


src/images/folder-disabled.svg
```diff
@@ -16,7 +16,6 @@
    sodipodi:version="0.32"
    inkscape:version="0.91 r13725"
    sodipodi:docname="folder-disabled.svg"
-   inkscape:export-filename="/home/jimmac/Desktop/horlander-style3.png"
```
NotMyFault, line 19: Instead of editing the icons here and adding them again in `/webapp/images/`, they seem good to be moved over, alongside removing the PNG icons.

For the record, this would address https://issues.jenkins.io/browse/JENKINS-66698

daeho-ro:
The folder icon is not changed in the jenkins 2.321 when I set the icon size.
This pr should be merged and I really waiting for this. 👍

src/images/folder-disabled.svg
zbynek: These are different files: the ones in `webapp` are optimized using SVGO's [UI](https://jakearchibald.github.io/svgomg/). The ones in main source are mostly unchanged, I only removed metadata specific to author's computer.

Not sure what's the best long term solution here: we could put only the optimized SVGs in the repo, or hook SVGO into the build process. There was some discussion about this in Jenkins core and IIRC the preference was for the latter, but that would require too many changes for this small PR, so for now I just kept both.

zbynek:
Tagging jenkinsci/jenkins#5778 for "on hold" reference.

timja (approved):
@fcojfernandez can you take a look please?

fcojfernandez (approved):
Seems ok, not tested so far. I will give some time in case anyone from @jenkinsci/cloudbees-folder-plugin-developers is interested in review it.

NotMyFault:
@fcojfernandez Thanks for merging! Could you attempt to trigger a release again? The build doing the release was aborted and superseded by the commit returning to snapshots.
The [artifactory](https://repo.jenkins-ci.org/releases/org/jenkins-ci/plugins/cloudbees-folder/) is still stuck on 6.16.

fcojfernandez:
@NotMyFault release out now. I had a problem when authenticating into artifactory, and I needed to fix it first

NotMyFault:
> 

Aight, thanks for resolving it quickly 👏🏻

jglick:
> I had a problem when authenticating into artifactory

Consider https://www.jenkins.io/doc/developer/publishing/releasing-cd/

PR #204, opened 2022-01-01
introduced diamond operator, use try with resources, minor improvements

StefanSpieker:
* introduced diamond operator, 
* use try with resources, 
* minor improvements

* Internal

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
@@ -505,7 +505,7 @@ public class FolderCredentialsProvider extends CredentialsProvider {
                 if (list == null || list.isEmpty()) {
                     return Collections.emptyList();
                 }
-                return Collections.unmodifiableList(list);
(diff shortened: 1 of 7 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
```diff
@@ -505,7 +505,7 @@ private synchronized boolean addCredentials(@NonNull Domain domain, @NonNull Cre
                 if (list == null || list.isEmpty()) {
                     return Collections.emptyList();
                 }
-                return Collections.unmodifiableList(new ArrayList<Credentials>(list));
+                return Collections.unmodifiableList(list);
```
jtnord, line 508: on the surface this seems wrong as the list returned my `getDomainCredentialsMap().get(domain);` may not be immutable, thus the unmodifiable collection could actually be modified after it was returned.

```suggestion
                return Collections.unmodifiableList(new ArrayList<>(list));
```

jtnord (changes requested):

jtnord (approved):

PR #207, opened 2022-01-31
Extend startup notification timeout as jobs continue to be loaded

basil:
**Untested** PR to consume the new API introduced in jenkinsci/jenkins#6237 to keep pushing back the startup timeout as long as Jenkins is continuing to make progress in loading jobs. This should make it far less likely that people will hit the `systemd` service startup timeout when using `systemd` with `Type=notify`.

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
@@ -40,6 +40,7 @@ import static hudson.Util.fixEmpty;
 import hudson.XmlFile;
 import hudson.init.InitMilestone;
 import hudson.init.Initializer;
+import hudson.lifecycle.Lifecycle;
 import hudson.model.AbstractItem;
 import hudson.model.Action;
 import hudson.model.AllView;
@@ -82,6 +83,9 @@ import hudson.views.ViewsTabBar;
 import java.io.File;
 import java.io.FileFilter;
 import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -586,6 +600,38 @@ public String call(I item) {
         }
     }
 
+    private static void onExtendTimeout(long timeout, @NonNull TimeUnit unit) {
+        // TODO When baseline is updated to jenkinsci/jenkins#6237, use Jenkins.get().getLifecycle().onExtendTimeout(timeout, unit)
```
jtnord, line 604:
```suggestion
        // TODO When baseline is updated to Jenkins-2.335 jenkinsci/jenkins#6237, use Jenkins.get().getLifecycle().onExtendTimeout(timeout, unit)
```

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -565,8 +570,17 @@ public String call(I item) {
                     float percentage = 100.0f * jobEncountered.incrementAndGet() / Math.max(1, jobTotal.get());
                     long now = System.currentTimeMillis();
                     if (loadingTick == 0) {
+                        // Buy ourselves two tick intervals to complete the first tick + some slop.
+                        onExtendTimeout(2 * TICK_INTERVAL, TimeUnit.MILLISECONDS);
```
jtnord, line 574: can throw exceptions - which would completely break loading here.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -586,6 +600,38 @@ public String call(I item) {
...
+            onExtendTimeout.invoke(lifecycle, timeout, unit);
+        } catch (IllegalAccessException e) {
+            IllegalAccessError x = new IllegalAccessError(e.getMessage());
+            x.initCause(e);
+            throw x;
+        } catch (InvocationTargetException e) {
+            Throwable t = e.getCause();
+            if (t instanceof RuntimeException) {
+                throw (RuntimeException) t;
+            } else if (t instanceof IOException) {
```
jtnord, line 623: is not documented to be thrown in https://github.com/jenkinsci/jenkins/blob/0c4b79128d1c437955f62d8c4332301f28bb8615/core/src/main/java/hudson/lifecycle/Lifecycle.java#L309 how can this happen?

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -586,6 +600,38 @@ public String call(I item) {
...
+                throw new UncheckedIOException((IOException) t);
+            } else if (t instanceof Exception) {
```
jtnord, line 625: same as above?

basil:
Closing in favor of #220.

PR #211, opened 2022-02-15
Enable JEP-229 CD

jtnord:
See [JENKINS-XXXXX](https://issues.jenkins-ci.org/browse/JENKINS-XXXXX).

* ...

### Submitter checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,7 @@
   </parent>
 
   <artifactId>cloudbees-folder</artifactId>
-  <version>${revision}${changelist}</version>
+  <version>6.${changelist}</version>
   <packaging>hpi</packaging>
 
   <name>Folders Plugin</name>
@@ -19,8 +19,7 @@
   </description>
   <url>https://github.com/jenkinsci/cloudbees-folder-plugin/</url>
   <properties>
-    <revision>6.18</revision>
-    <changelist>-SNAPSHOT</changelist>
+    <changelist>999999-SNAPSHOT</changelist>
     <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
     <jenkins.version>2.222.4</jenkins.version>
     <java.level>8</java.level>
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+@jenkinsci/cloudbees-folder-plugin-developers
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,10 @@
 version: 2
 updates:
-  - package-ecosystem: "maven"
-    directory: "/"
-    schedule:
-      interval: "weekly"
+- package-ecosystem: "maven"
+  directory: "/"
+  schedule:
+    interval: "weekly"
+- package-ecosystem: github-actions
+  directory: /
+  schedule:
+    interval: weekly
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,4 +1,1 @@
 _extends: .github
-version-template: $MAJOR.$MINOR.$PATCH
-tag-template: cloudbees-folder-$NEXT_PATCH_VERSION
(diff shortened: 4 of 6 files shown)


car-roll (approved):

olamy:
i would prefer using a solution provided by Jenkins tooling rather than a concurrent tool. 
Can we still release manually?

jtnord:
> i would prefer using a solution provided by Jenkins tooling rather than a concurrent tool.

the CD tooling is provided by the Jenkins community - not sure what you mean by concurrent?

> Can we still release manually?

until your user has deploy permissions removed yes (which is not something happening at the moment for CD enabled repos) but you should not even attempt to do that once CD is enabled save for security issues - which do not use CD anyway.

olamy:
> > i would prefer using a solution provided by Jenkins tooling rather than a concurrent tool.
> 
> 
> 
> the CD tooling is provided by the Jenkins community - not sure what you mean by concurrent?
> 

But based on a concurrent tool called GitHub actions..
I'm pretty sure and I really hope as a jenkins contributor Jenkins can do the for his community 

> 
> 
> > Can we still release manually?
> 
> 
> 
> until your user has deploy permissions removed yes (which is not something happening at the moment for CD enabled repos) but you should not even attempt to do that once CD is enabled save for security issues - which do not use CD anyway.
> 
> 
> 

CD or not is not the point of my comment

pom.xml
```diff
@@ -10,7 +10,7 @@
   </parent>
 
   <artifactId>cloudbees-folder</artifactId>
-  <version>${revision}${changelist}</version>
+  <version>6.${changelist}</version>
```
jglick, line 13: or
```suggestion
  <version>${changelist}</version>
```

CC @daniel-beck

jglick (approved):

jglick:
> i would prefer using a solution provided by Jenkins tooling

Well, we do not have one, and due to the way permissions are managed in this org it would likely be difficult to set up, such as a separate server running the `github-oauth` plugin. What we have is pretty simple and works fine (remember ci.jenkins.io is still used for CI / testing) so there is little incentive for infrastructure folks to invest a lot of time creating something more complicated to accomplish the same thing.

jglick:
Merge conflict due to NLs in f6905c07c8e1d3ca204f193062fd1fd2f253cbf5 (https://issues.apache.org/jira/browse/MRELEASE-899), ironically.

olamy:
> Merge conflict due to NLs in [f6905c0](https://github.com/jenkinsci/cloudbees-folder-plugin/commit/f6905c07c8e1d3ca204f193062fd1fd2f253cbf5) (https://issues.apache.org/jira/browse/MRELEASE-899), ironically.

**AGAIN**  I do not have any concern about CD or not using the maven-release-plugin. I'm perfectly fine with that. That's why I don't understand why it's ironic? 
I just don't understand why the Jenkins community would not use some tooling provided by Jenkins but rather use some tooling based on concurrent tool such GitHub action.  
But it has been approved by the Jenkins board, it's a democratic process and I can/will have to live with that. 
Note I do no reject or request changes for this PR but as it's supposed to be a democratic process so at least I express my point of view.  (period)

olamy:
> > i would prefer using a solution provided by Jenkins tooling
> 
> Well, we do not have one, and due to the way permissions are managed in this org it would likely be difficult to set up, such as a separate server running the `github-oauth` plugin. What we have is pretty simple and works fine (remember ci.jenkins.io is still used for CI / testing) so there is little incentive for infrastructure folks to invest a lot of time creating something more complicated to accomplish the same thing.

I have proposed implementing this ("using our own dog food rather than concurrent tool") as a possible GSoC (https://community.jenkins.io/t/google-summer-of-code-gsoc-2022-call-for-organizers-mentors-and-project-ideas/1010/26).
This can maybe attract some attention. At the end we are talking about running 4/5 command lines, I hope we can implement this via some Jenkins tooling and Jenkins community infrastructure.
This can be a very good subject for some intern to learn working on a project which will need to work some communication skills as well with different teams.

PR #218, opened 2022-03-01
[JENKINS-67921] Add a shutdown hook stopping all ongoing folder computations

Vlatombe:
When running JenkinsRule based tests on windows, there are often locks on indexing.log
which is the file holding the computation logs.

This means a folder computation is running while Jenkins is being
shutdown, and this computation is not aborted before shutting down.

See [JENKINS-67921](https://issues.jenkins-ci.org/browse/JENKINS-67921).

* Entry 1: Shutdown ongoing folder computations when shutting down Jenkins
* ...

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
@@ -28,6 +28,7 @@ import com.cloudbees.hudson.plugins.folder.AbstractFolder;
 import com.thoughtworks.xstream.XStreamException;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import hudson.Extension;
 import hudson.ExtensionList;
 import hudson.Util;
 import hudson.XmlFile;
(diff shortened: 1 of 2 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
```diff
@@ -1121,4 +1121,18 @@ public void completed(String name) {
          */
         NO_RECALCULATION
     }
+
+    @Extension
+    public static class StopComputationsOnShutdown extends ItemListener {
+        @Override
+        public void onBeforeShutdown() {
+            for (ComputedFolder cf : Jenkins.get().getAllItems(ComputedFolder.class)) {
```
jglick, line 1129:
```suggestion
            for (ComputedFolder cf : Jenkins.get().allItems(ComputedFolder.class)) {
```

src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder2Test.java
```diff
@@ -76,6 +77,20 @@ public void evaluate() throws Throwable {
         });
     }
 
+    @Test
```
jglick, line 80:
```suggestion
    @Issue("JENKINS-67921")
    @Test
```

jglick (approved):

amuniz (approved):
:+1:

jglick:
Whether or not you found another solution to your test failures, this still seems like a legitimate bug fix to me, right?

Vlatombe:
In theory yes but in practice this would be only useful during tests since either the JVM shutdowns, or eventually the branch indexing thread completes.

jglick:
> either the JVM shutdowns

Meaning the process exits and releases any locks. OK.

PR #222, opened 2022-03-19
List credentials authorisation check

ChampiYann:
There is a [comment in this Github issue](https://github.com/jenkinsci/ghprb-plugin/issues/534#issuecomment-559539250) detailing the problem this pull requests tries to fix.
It comes down to adding a check to the getCredentials method to check if the user requesting the credentials is permitted to use the credentials of a provider.

* ...

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
@@ -122,6 +122,21 @@ public class FolderCredentialsProvider extends CredentialsProvider {
                                                           @NonNull List<DomainRequirement> domainRequirements) {
         List<C> result = new ArrayList<>();
         Set<String> ids = new HashSet<>();
+
+        // // Check if Authentication is authorised to list the credentials
+        // Boolean isAuthorised = false;
+        // if (ACL.SYSTEM.equals(authentication)) {
+        //     isAuthorised = true;
+        // }
+        // if (itemGroup instanceof AbstractFolder) { 
+        //     final AbstractFolder<?> folder = AbstractFolder.class.cast(itemGroup);
+        //     if (folder.hasPermission(authentication, CredentialsProvider.USE_ITEM)) {
+        //         isAuthorised = true;
+        //     }
+        // }
+        // // Get credentials
+        // if (Boolean.TRUE.equals(isAuthorised)) {
+
         if (ACL.SYSTEM.equals(authentication)) {
             while (itemGroup != null) {
                 if (itemGroup instanceof AbstractFolder) {
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
@@ -136,8 +136,6 @@ public class FolderCredentialsProvider extends CredentialsProvider {
         }
         // Get credentials
         if (Boolean.TRUE.equals(isAuthorised)) {
-
-        // if (ACL.SYSTEM.equals(authentication)) {
             while (itemGroup != null) {
                 if (itemGroup instanceof AbstractFolder) {
                     final AbstractFolder<?> folder = AbstractFolder.class.cast(itemGroup);
@@ -207,8 +205,6 @@ public class FolderCredentialsProvider extends CredentialsProvider {
         }
         // Get credentials
         if (Boolean.TRUE.equals(isAuthorised)) {
-        
-        // if (ACL.SYSTEM.equals(authentication)) {
             while (itemGroup != null) {
                 if (itemGroup instanceof AbstractFolder) {
                     final AbstractFolder<?> folder = AbstractFolder.class.cast(itemGroup);
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
@@ -123,21 +123,21 @@ public class FolderCredentialsProvider extends CredentialsProvider {
         List<C> result = new ArrayList<>();
         Set<String> ids = new HashSet<>();
 
-        // // Check if Authentication is authorised to list the credentials
-        // Boolean isAuthorised = false;
-        // if (ACL.SYSTEM.equals(authentication)) {
-        //     isAuthorised = true;
-        // }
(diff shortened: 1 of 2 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
```diff
@@ -178,7 +191,20 @@
...
         Set<String> ids = new HashSet<>();
+
+        // Check if Authentication is authorised to list the credentials
+        Boolean isAuthorised = false;
         if (ACL.SYSTEM.equals(authentication)) {
+            isAuthorised = true;
+        }
+        if (itemGroup instanceof AbstractFolder) { 
+            final AbstractFolder<?> folder = AbstractFolder.class.cast(itemGroup);
+            if (folder.hasPermission(authentication, CredentialsProvider.USE_ITEM)) {
```
jtnord, line 202: I think this should allow for listing credentials if the user can also configure the itemgroup (as they may be able to select a credential that the system (not they themselves) will use later

(note the follow syntax has not been checked)

```suggestion
            if (folder.hasPermission(authentication, CredentialsProvider.USE_ITEM) || folder.hasPermission(authentication, Item.CONFIGURE)) {
```

@daniel-beck  I always have to double think this - but for listing the credential IDs it should be you have configure on the item, or the ability to use a credntials.  (configure so you can select a credential that the job/system can use even if you can not use it, USE_ITEM incase you can not configure the job (e.g. for the pipeline snippet generator at the job level and the pipeline is "as-code")?

src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
```diff
@@ -122,7 +122,20 @@
                                                           @NonNull List<DomainRequirement> domainRequirements) {
         List<C> result = new ArrayList<>();
         Set<String> ids = new HashSet<>();
+
+        // Check if Authentication is authorised to list the credentials
+        Boolean isAuthorised = false;
         if (ACL.SYSTEM.equals(authentication)) {
+            isAuthorised = true;
+        }
+        if (itemGroup instanceof AbstractFolder) {
```
jtnord, line 131:
```suggestion
        } else if (itemGroup instanceof AbstractFolder) { 
```

src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
```diff
@@ -122,7 +122,20 @@
                                                           @NonNull List<DomainRequirement> domainRequirements) {
         List<C> result = new ArrayList<>();
         Set<String> ids = new HashSet<>();
+
+        // Check if Authentication is authorised to list the credentials
+        Boolean isAuthorised = false;
```
jtnord, line 127: prefer primatives 

```suggestion
        boolean isAuthorised = false;
```

src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
```diff
@@ -122,7 +122,20 @@
...
+            isAuthorised = true;
+        }
+        if (itemGroup instanceof AbstractFolder) { 
+            final AbstractFolder<?> folder = AbstractFolder.class.cast(itemGroup);
+            if (folder.hasPermission(authentication, CredentialsProvider.USE_ITEM)) {
+                isAuthorised = true;
+            }
+        }
+        // Get credentials
+        if (Boolean.TRUE.equals(isAuthorised)) {
```
jtnord, line 138:
```suggestion
        if (isAuthorised) {
```

src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
```diff
@@ -178,7 +191,20 @@
                                                                    @NonNull CredentialsMatcher matcher) {
         ListBoxModel result = new ListBoxModel();
         Set<String> ids = new HashSet<>();
+
+        // Check if Authentication is authorised to list the credentials
+        Boolean isAuthorised = false;
         if (ACL.SYSTEM.equals(authentication)) {
+            isAuthorised = true;
+        }
+        if (itemGroup instanceof AbstractFolder) {
```
jtnord, line 200:
```suggestion
        else if (itemGroup instanceof AbstractFolder) { 
```

src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
```diff
@@ -178,7 +191,20 @@
                                                                    @NonNull CredentialsMatcher matcher) {
         ListBoxModel result = new ListBoxModel();
         Set<String> ids = new HashSet<>();
+
+        // Check if Authentication is authorised to list the credentials
+        Boolean isAuthorised = false;
```
jtnord, line 196:
```suggestion
        boolean isAuthorised = false;
```

src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
```diff
@@ -178,7 +191,20 @@
...
+            isAuthorised = true;
+        }
+        if (itemGroup instanceof AbstractFolder) { 
+            final AbstractFolder<?> folder = AbstractFolder.class.cast(itemGroup);
+            if (folder.hasPermission(authentication, CredentialsProvider.USE_ITEM)) {
+                isAuthorised = true;
+            }
+        }
+        // Get credentials
+        if (Boolean.TRUE.equals(isAuthorised)) {
```
jtnord, line 207:
```suggestion
        if (isAuthorised) {
```

jtnord (changes requested):

src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
jtnord: which I would agree with if not for https://github.com/jenkinsci/cloudbees-folder-plugin/pull/222#discussion_r831064549

src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
ChampiYann: Hi, I'm assuming we're waiting for @daniel-beck 's response on this. Is there any indication on when we can expect that?

PR #227, opened 2022-04-04
Adding missing attribute for indexing interval

nirkoren:
See [JENKINS-XXXXX](https://issues.jenkins-ci.org/browse/JENKINS-XXXXX).

* ...
Adding additional entry for polling 

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/computed/PeriodicFolderTrigger.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/PeriodicFolderTrigger.java
@@ -263,6 +263,7 @@ public class PeriodicFolderTrigger extends Trigger<ComputedFolder<?>> {
             ListBoxModel model = new ListBoxModel();
             model.add("1 minute", "1m");
             model.add("2 minutes", "2m");
+            model.add("3 minutes", "3m");
             model.add("5 minutes", "5m");
             model.add("10 minutes", "10m");
             model.add("15 minutes", "15m");


nirkoren:
> 

Thanks a lot, I went over the code and seems that there is no reason why not providing it. We are using 3 minutes in the multibranch pipeline DSL (using interval 18000) and it works in fact but we cannot see it in the UI. This is the reason I want to add this change

jglick:
Unfortunately AFAIK Jenkins lacks a standard form control for https://en.wikipedia.org/wiki/Combo_box (CC @timja).

timja:
> Unfortunately AFAIK Jenkins lacks a standard form control for [en.wikipedia.org/wiki/Combo_box](https://en.wikipedia.org/wiki/Combo_box) (CC @timja).

`<f:combobox />` ?

https://weekly.ci.jenkins.io/design-library/TextBox/

jglick:
Huh, forgot about that, though apparently I once knew: https://github.com/jenkinsci/jenkins/pull/1510

Might then suffice to just update https://github.com/jenkinsci/cloudbees-folder-plugin/blob/79e858ef76a2c6464be5ccd6dca72e5af5a5c408/src/main/resources/com/cloudbees/hudson/plugins/folder/computed/PeriodicFolderTrigger/config.jelly#L27

nirkoren:
Hi @jglick: Thanks a lot! Can you please let me know what is missing in order to approve my pull request?

jglick:
I am not a maintainer and would not approve, just offering suggestions. From what I recall of this field and its usage, I would suggest reverting the new option (no need for it), switching to a combo box, offering the current values as suggestions, and adding form validation to verify syntax in case the user types in some other value.

nirkoren:
@jtnord  - are you the maintainer?

timja:
Maintainers (including @jglick):
image

nirkoren:
So, who should accept / decline this PR?

alecharp (approved):
I don't see any harm in adding this to the plugin.

PR #232, opened 2022-04-21
Update plugin parent POM and BOM

basil:


basil:
@jenkinsci/cloudbees-folder-plugin-developers A release of this would facilitate Java 17 PCT testing.

pom.xml
```diff
@@ -58,8 +57,8 @@
     <dependencies>
       <dependency>
         <groupId>io.jenkins.tools.bom</groupId>
-        <artifactId>bom-2.277.x</artifactId>
-        <version>984.vb5eaac999a7e</version>
+        <artifactId>bom-2.289.x</artifactId>
+        <version>1289.v5c4b_1c43511b_</version>
```
jtnord, line 61: @basil I am generally against bumping bom versions unless it is needed to pick up a fix specifically needed for the plugin as it causes issues when upgrading for a fix.  as the PCT will test with newer versions and this plugin is part of the bom - is this strictly needed give the 984 version is availabe for the 2.289 bom?

jglick: > it causes issues when upgrading for a fix

Can you elaborate?

jtnord: example 
plugin A uses the bom, has dependencies on B C D E.....

bom update is merged to A.  (uninteresting so no CD release is made)

bug fix is made to A (new release is made by CD)

to get the bug fix you now need to upgrade A B C D and E....  

this causes issues as B C D E whilst they may have past unit tests, they may contain bugs that are not found.
it also makes it harder to get users to test fixes as they need to upgrade way more than just a plugin.

regressions do to the upgrade are harder to pin down, `git bisect` will see just one commit upgrading the world, so you need to resort to non standard scripts to get and test the versions of the bom between last release and now.

TL;DR if the plugin does not need an upgrade of B C D or E then there is no reason to upgrade the bom especially if the plugin is itself part of the bom as the compatability testing will happen in the BOM (at least binary compatability, not currently source, but if that is cared about it is a simple thing to do)

jtnord: for this plugin it *only* affect credentials (others are test scope), but generally having special case plugins is root to pain.

jglick: > B C D E whilst they may have past unit tests, they may contain bugs that are not found

Of course they may, but these are long since on the update center in users’ hands, which would be the more important consideration by far.

> regressions do to the upgrade are harder to pin down

You mean regressions in tests? While it is possible for a BOM update to regress production functionality, that would be rare.

> harder to get users to test fixes

Regardless of how or even if you are using the BOM, in general if you are asking someone to test a particular fix in situ, it is advisable to ask which version they are currently running, then create a draft PR rooted at that version (or the subsequent commit, if still using MRP) backporting the fix with `git cherry-pick -x -m1`, then point them to the resulting incremental build.

jtnord: > Regardless of how or even if you are using the BOM, in general if you are asking someone to test a particular fix in situ,

this is creating extra work when they are already on the latest release.

for users how are upgrading updating the bom in the plugin gets them nothing - it is a developer thing.
for users that are not upgrading biumping the bom unesescarily is creating more work for the developer and the user.

jtnord: >  but these are long since on the update center in users’ hands, which would be the more important consideration by far.

isn;t the bom updated and released by CD and dependabot so long since is actually a day?

jglick: > isn’t the bom updated and released by CD

Dependabot bumps to `bom` are marked `dependencies` so they are not automatically released. Typically we only bother to cut a release when adding new plugins, or doing upgrades which fix some sort of logistical POM issue.

basil: > as the PCT will test with newer versions

[…]

> especially if the plugin is itself part of the bom as the compatability testing will happen in the BOM

https://github.com/jenkinsci/bom/issues/821

basil:
Is there a reason https://github.com/jenkinsci/cloudbees-folder-plugin/pull/232#pullrequestreview-948466391 has not been dismissed? As demonstrated in https://github.com/jenkinsci/cloudbees-folder-plugin/pull/232#discussion_r855331909 the premise of the review was flawed; therefore, the conclusion is flawed.

pom.xml
jglick: (of course we expect that to be fixed at some point)

jglick (approved):
(FWIW)

basil:
@jtnord ?

jtnord:
> Is there a reason [#232 (review)](https://github.com/jenkinsci/cloudbees-folder-plugin/pull/232#pullrequestreview-948466391) has not been dismissed? As demonstrated in [#232 (comment)](https://github.com/jenkinsci/cloudbees-folder-plugin/pull/232#discussion_r855331909) the premise of the review was flawed; therefore, the conclusion is flawed.

because there is absolutely nothing wrong with https://github.com/jenkinsci/cloudbees-folder-plugin/pull/232#discussion_r855142567 ?  the only part that is flawed is the assumption that the BOM is actually testing things in combination.  as it is not - this makes me *more* wary to pickup boms as there has not been the testing that I explicitly thought there had boon :-o

jglick:
If the BOM update is controversial, can the rest be refiled?

basil:
It can. I have no plans to do so.

jtnord:
> If the BOM update is controversial, can the rest be refiled?

I left an online suggestion.

basil:
I do not accept your suggestion.

basil:
Since I wrote the above, James contacted me to explain that dependency updates sometimes cause him pain when dealing with ATH failures. Is an ATH maintainer who is experiencing pain caused by dependency updates justified in blocking a pull request that updates dependencies and could possibly cause them more pain? Not if it is uncertain that the pull request will _actually_ cause the ATH maintainer pain. In fact, in this case the BOM maintainer could be justified in updating dependencies in order to mitigate their _own_ pain resulting from https://github.com/jenkinsci/bom/issues/821. So here we have a conflict between two maintainers, each of whom is justified in their actions from their own perspective but neither of whom deserves the conflict. In order to eliminate the conflict at its root and optimize globally rather than locally, we need to shift testing left in the software delivery lifecycle. I have started a [mailing list thread](https://groups.google.com/g/jenkinsci-dev/c/OmM5-f-ohtI/m/wYzeKH2BAgAJ) to discuss this in a broader forum.

PR #239, opened 2022-06-03
Icon for folder credentials store is gone from credentials plugin

twasyl:
See [JENKINS-68674](https://issues.jenkins-ci.org/browse/JENKINS-68674).

Before this PR, the icon for the folder credentials in the crumb bar was missing. Now with the change, the icon is back:
folder-icon

* Entry 1: Issue, Icon for folder credentials store has been updated to match credentials plugin change

### Submitter checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
     <changelist>999999-SNAPSHOT</changelist>
     <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
     <!-- remember to change the io.jenkins.tools.bom artifact when changing this -->
-    <jenkins.version>2.346.1</jenkins.version>
+    <jenkins.version>2.346</jenkins.version>
     <no-test-jar>false</no-test-jar>
     <hpi.compatibleSinceVersion>5.2</hpi.compatibleSinceVersion>
   </properties>
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
     <changelist>999999-SNAPSHOT</changelist>
     <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
     <!-- remember to change the io.jenkins.tools.bom artifact when changing this -->
-    <jenkins.version>2.346</jenkins.version>
+    <jenkins.version>2.319.1</jenkins.version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -22,7 +22,7 @@
     <changelist>999999-SNAPSHOT</changelist>
     <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
     <!-- remember to change the io.jenkins.tools.bom artifact when changing this -->
-    <jenkins.version>2.277.1</jenkins.version>
+    <jenkins.version>2.346.1</jenkins.version>
```
jglick, line 25: Did you mean 
```suggestion
    <jenkins.version>2.346</jenkins.version>
```
?

jglick: How could you have tried this PR if the baseline here does not even exist yet? Did you amend the POM afterwards?

jglick:
Also please consider my note in Jira that the icon appears to be used only in this plugin (confirm!) so it would make sense to move it here from `credentials`, along with the icon set registration code for it.

pom.xml
jglick: And why the very new baseline? 2.319.x should suffice for the fix I think? https://github.com/jenkinsci/credentials-plugin/pull/291

twasyl:
> I believe the icon is located in credentials

Yes it is:

filetree

twasyl:
@jglick I made the icon migration

icon-migration

jglick (approved):
Looks right.

jglick:
Fix released. Can you file a cleanup PR in `credentials`?

papemamadou91:
Hello,

I think it is not a good idea to remove this icon from credentials plugin.
Since, we updated our credentials plugin to their latest version, we have some icon that are missing (see picture below)

image


And after inspecting the code, we see that it is the image **folder-store.svg** which the credentials plugin need, that is missing 

image



**Our actual configuration** 
Jenkins : 2.346.1
Credentials plugin : 1139.veb_9579fca_33b_  (this is the latest)
Cloudbees-Folder Plugin : 6.729.v2b_9d1a_74d673 (this is the latest)

jglick:
⇒ #253

PR #264, opened 2022-09-19
Implement `TriggeredItem` from `ComputedFolder`

jglick:
Downstream of https://github.com/jenkinsci/jenkins/pull/7131. Cleans up tech debt introduced in #20 (amended slightly in #119).


julieheard (approved):
Looks good to me

olamy (approved):

basil:
Incompatible; needs a consumer to be adapted as described in https://github.com/jenkinsci/jenkins/pull/7244#pullrequestreview-1136272015.

riywo:
FYI: This change starts causing our integration test failure as the current LTS version of Jenkins is 2.361.2 https://www.jenkins.io/download/

https://github.com/jenkinsci/autify-plugin/actions/runs/3212345732/jobs/5251131458#step:4:1522

We're going to switch to the latest Jenkins for our integration test, so it's ok for us. But the user who uses the Jenkins LTS version might face the same issue on their production.

jglick:
> the user who uses the Jenkins LTS version might face the same issue

No, because they would not be offered this update to begin with. The problem lies in your integration test script.

riywo:
Got it, thank you for your confirmation.

Just FYI: We use this tool to install the required plugins https://github.com/jenkinsci/plugin-installation-manager-tool

https://github.com/jenkinsci/autify-plugin/blob/main/integration-test/integration-test.bash#L55

jglick:
> `--jenkins-version`: (optional) Version of Jenkins to be used. If not specified, the plugin manager will try to extract it from the WAR file or other sources. The argument can be also set using the `JENKINS_VERSION` environment variable.

Without specifying the Jenkins version, or allowing it to be inferred, PIMT will just download the latest version of every plugin, whether it is compatible or not.

riywo:
I thought it would take the version from the WAR file specified by `--war`. Probably we should set the correct path for the WAR file because cwd is different. Thank you for pointing that.

PR #266, opened 2022-10-14
Rewrite of `PeriodicFolderTrigger.run`

jglick:
The logic of `PeriodicFolderTrigger.run` was very difficult to understand (not helped by the lack of clear logging), did not quite match the behavior documented in `help.html` and implied by its label **Periodically if not otherwise run**, and brought in a weird mixture of `transient` and `static` implicit state not present in `indexing.xml`, not to mention `Random`.

The [original problem](https://issues.jenkins.io/browse/JENKINS-42511?focusedCommentId=290633&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-290633) was that under some circumstances, `FolderCron` (plain `Trigger.Cron` as of #264) would run several triggered minute slots in close succession, which if you were unlucky could result in multiple simultaneous (clashing) folder indexing runs. #91 tried to solve that by introducing `lastTriggered`, but in fact it seems easier to just delay the folder indexing by a few seconds and let the queue handle coalescing. You can actually see this working right after controller restart if you have a short interval like 1m.

[Another problem](https://issues.jenkins.io/browse/JENKINS-43424) was that you do not want lots of different computed folders to index at once. #99 addressed that with `H/*` syntax, but also introduced a randomized delay after `startup` which seems to have only been needed because we were already relying on `lastTriggered` rather than actually checking the timestamp on disk. This patch keeps the balanced schedule fix.

#119 addressed a problem with a short interval like 1m: since cron only runs every minute, it could decide that the last indexing was just a little bit too recently, and skip it until the next minute slot. It is necessary to adjust the interval downward a little to accommodate that; the adjustment is tricky because the crontab spec also does not exactly match the interval in general (for longer intervals we do not waste CPU on `run` every minute). This patch keeps the same proportional logic, with an additional fixed offset.

This patch should handle all three issues, to the extent I can tell without test coverage (🎶 _if I could mock out time, if I could find a way_), with a much simpler and more intuitive implementation that also behaves uniformly across controller restarts since it is just driven by the visible and persisted timestamp.

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/computed/PeriodicFolderTrigger.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/PeriodicFolderTrigger.java
@@ -26,15 +26,14 @@ package com.cloudbees.hudson.plugins.folder.computed;
 import antlr.ANTLRException;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.Extension;
-import hudson.model.Cause;
+import hudson.Util;
 import hudson.model.Item;
 import hudson.model.Items;
 import hudson.triggers.TimerTrigger;
 import hudson.triggers.Trigger;
 import hudson.triggers.TriggerDescriptor;
 import hudson.util.ListBoxModel;
-import java.util.List;
-import java.util.Random;
+import java.util.Date;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/computed/PeriodicFolderTrigger.java
```diff
@@ -184,50 +169,34 @@ public String getInterval() {
     /**
      * {@inheritDoc}
      */
-    @SuppressWarnings("ConstantConditions")
-    @SuppressFBWarnings("RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE")
+    @SuppressFBWarnings(value = {"RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE", "NP_NULL_ON_SOME_PATH"}, justification = "ComputedFolder.computation is set in init but that is overridable so let us just make sure; Trigger.job once set is never cleared")
```
jtnord, line 172: ❤️

jtnord:
> 🎶 if I could mock out time, if I could find a way

just a small write to the Jenkins project to use `Clock` everywhere!

jtnord (approved):

omehegan:
🙇 so great to see some improvements in this area!

PR #268, opened 2022-10-21
[JENKINS-69914] Remove overcomplicated `checkUrl` (CSP compatibility)

Pldi23:
Hello, HacktoberFest is here ([Content Security Policy: smooth introduction Topic](https://issues.jenkins.io/browse/JENKINS-60865))

See [JENKINS-69914](https://issues.jenkins.io/browse/JENKINS-69914).

[Content-Security-Policy Compatibility documentation](https://www.jenkins.io/doc/developer/security/csp/#legacy-javascript-checkurl-validation)

Manual testing recoding:

https://user-images.githubusercontent.com/47667812/197193449-126ba19b-fd61-4150-884f-b665da66c3e9.mov

### Submitter checklist

- [X] Tested manually ~~Appropriate autotests or explanation to why this change has no tests~~

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
@@ -30,6 +30,7 @@ import hudson.model.DescriptorVisibilityFilter;
 import hudson.model.Item;
 import hudson.model.TopLevelItem;
 import hudson.model.TopLevelItemDescriptor;
+import hudson.util.FormValidation;
 import hudson.views.ViewsTabBar;
 import java.io.File;
 import java.util.ArrayList;
@@ -38,6 +39,8 @@ import edu.umd.cs.findbugs.annotations.CheckForNull;
 import jenkins.model.Jenkins;
 import jenkins.model.ProjectNamingStrategy;
 import org.jenkins.ui.icon.IconSpec;
+import org.kohsuke.stapler.AncestorInPath;
+import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.Stapler;
 import org.kohsuke.stapler.StaplerRequest;
 
@@ -128,6 +131,11 @@ public abstract class AbstractFolderDescriptor extends TopLevelItemDescriptor im
         }
         return r;
     }
+
+    public FormValidation doCheckDisplayNameOrNull(@AncestorInPath AbstractFolder folder, @QueryParameter String value) {
+        return Jenkins.get().doCheckDisplayName(value, folder.getName());
+    }
+
 
     /**
      * Needed if it wants Folder are categorized in Jenkins 2.x.
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
@@ -39,6 +39,8 @@ import edu.umd.cs.findbugs.annotations.CheckForNull;
 import jenkins.model.Jenkins;
 import jenkins.model.ProjectNamingStrategy;
 import org.jenkins.ui.icon.IconSpec;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.AncestorInPath;
 import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.Stapler;
@@ -132,6 +134,7 @@ public abstract class AbstractFolderDescriptor extends TopLevelItemDescriptor im
         return r;
     }
 
+    @Restricted(NoExternalUse.class)
     public FormValidation doCheckDisplayNameOrNull(@AncestorInPath AbstractFolder folder, @QueryParameter String value) {
         return Jenkins.get().doCheckDisplayName(value, folder.getName());
     }
--- a/src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
+++ b/src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
@@ -66,7 +66,7 @@ THE SOFTWARE.
 
         <div class="jenkins-!-margin-top-6">
           <f:entry title="${%Display Name}" field="displayNameOrNull">
-            <f:textbox />
+            <f:textbox/>
           </f:entry>
         </div>
 
--- a/src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
+++ b/src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
@@ -66,7 +66,7 @@ THE SOFTWARE.
 
         <div class="jenkins-!-margin-top-6">
           <f:entry title="${%Display Name}" field="displayNameOrNull">
-            <f:textbox checkUrl="checkDisplayName" checkDependsOn=""/>
+            <f:textbox />
           </f:entry>
         </div>
 
--- a/src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
+++ b/src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
@@ -66,9 +66,6 @@ THE SOFTWARE.
 
         <div class="jenkins-!-margin-top-6">
           <f:entry title="${%Display Name}" field="displayNameOrNull">
-            <f:invisibleEntry>
-              <f:textbox name="jobName" value="${it.name}"/>
-            </f:invisibleEntry>
             <!--
             TODO Remove block-control class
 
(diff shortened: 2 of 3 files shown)


src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
```diff
@@ -66,6 +66,9 @@ THE SOFTWARE.
 
         <div class="jenkins-!-margin-top-6">
           <f:entry title="${%Display Name}" field="displayNameOrNull">
+            <f:invisibleEntry>
+              <f:textbox name="jobName" value="${it.name}"/>
+            </f:invisibleEntry>
```
jglick, line 71: I doubt you need this, or the `checkDependsOn`. (Or for that matter the `checkUrl`, provided the web method has the appropriate name: `doCheckDisplayNameOrNull`) Try replacing `@QueryParameter String jobName` with `@AncestorInPath AbstractFolder folder`.

jglick:
I took the liberty of adjusting the PR title to reflect the desired outcome, rather than the details of the implementation which are probably still overcomplicated.

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
Pldi23: Yes you're right AncestorInPath works well,  the only thing it's a little bit weird that we need empty `checkDependsOn=""`. Applied in [a3e7976](https://github.com/jenkinsci/cloudbees-folder-plugin/pull/268/commits/a3e7976633175a326562274397611b8dc7b91ee0)

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
```diff
@@ -78,7 +75,7 @@ THE SOFTWARE.
             The save/apply buttons are out of place until the DOM is modified by expanding config elements or by
             resizing the browser window.
             -->
-            <f:textbox clazz="block-control" checkUrl="checkDisplayName" checkDependsOn="jobName"/>
+            <f:textbox clazz="block-control" checkUrl="checkDisplayName" checkDependsOn=""/>
```
jglick, line 78:
```suggestion
            <f:textbox clazz="block-control" checkUrl="checkDisplayName"/>
```

or as previously mentioned simply
```suggestion
            <f:textbox clazz="block-control"/>
```

if you use the conventional check method name.

Also, this would be a good time to revisit the TODO comment above and maybe delete `clazz` as well if that is obsolete.

jglick (approved):

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
Pldi23: hm, I tried to apply both suggestions but for some reason they don't works

1. When I try `<f:textbox clazz="block-control" checkUrl="checkDisplayName"/>` the page becomes a blank screen. 
2. And when I try `<f:textbox clazz="block-control"/>` view displays well, but checkDisplayName request does not send

jglick: Not sure offhand, but form validation works fine in core and tons of plugins without overriding conventional values in attributes.

Pldi23: >Also, this would be a good time to revisit the TODO comment above and maybe delete clazz as well if that is obsolete.

Remove block-control class and tested on Chrome and Firefox

Pldi23: What I found, based on this [doc](https://www.jenkins.io/doc/developer/security/csp/#legacy-javascript-checkurl-validation)  `checkDependsOn` must exists but could be empty, as of Jenkins 2.360

jglick: I do not know why the doc says that, since there are tons of examples of form fields which implicitly use a `doCheckXXX` method without overriding either `checkUrl` or `checkDependsOn`, including some using `@AncestorInPath`.See https://github.com/jenkinsci/jenkins/blob/ccc0878ba856d18bb0fb9cffab61bfc69096efd1/core/src/main/java/jenkins/triggers/ReverseBuildTrigger.java#L209 + https://github.com/jenkinsci/jenkins/blob/ccc0878ba856d18bb0fb9cffab61bfc69096efd1/core/src/main/resources/jenkins/triggers/ReverseBuildTrigger/config.jelly#L28-L29 for one example; `TimerTrigger.spec` is another. I _suspect_ that your problem is that you are still using a nonstandard `checkUrl`. Try omitting it, and using the conventional check method name.

jglick: CC @daniel-beck as author of https://github.com/jenkins-infra/jenkins.io/pull/5301, which links to https://github.com/jenkinsci/jenkins/pull/6857 that looks much too complicated to me: https://github.com/jenkinsci/jenkins/blob/ccc0878ba856d18bb0fb9cffab61bfc69096efd1/core/src/main/java/hudson/slaves/SimpleScheduledRetentionStrategy.java#L261 is not using a conventional name `doCheckStartTimeSpec`; given that, I think https://github.com/jenkinsci/jenkins/blob/ccc0878ba856d18bb0fb9cffab61bfc69096efd1/core/src/main/resources/hudson/slaves/SimpleScheduledRetentionStrategy/config.jelly#L28-L31 could be greatly simplified to regular form binding (other than a custom `help` to avoid duplication)

```xml
<f:entry field="startTimeSpec" title="${%Startup Schedule}"  help="/descriptor/hudson.triggers.TimerTrigger/help/spec">
    <f:textarea/>
</f:entry>
```

Pldi23: Made it work by moving doCheck to descriptor. See [efce1f8](https://github.com/jenkinsci/cloudbees-folder-plugin/pull/268/commits/efce1f896631e15a5126151af3a606f26c742f14)

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
```diff
@@ -66,16 +66,7 @@ THE SOFTWARE.
...
-            TODO Remove block-control class
-
-            Having this forces application of the 2.x UI but it is glitchy if no f:optionalBlock or f:radioBlock
-            elements are in the form (such is the case for the standard Folder implementation).
-
-            The save/apply buttons are out of place until the DOM is modified by expanding config elements or by
-            resizing the browser window.
-            -->
-            <f:textbox clazz="block-control" checkUrl="'${rootURL}/checkDisplayName?displayName='+encodeURIComponent(this.value)+'&amp;jobName='+encodeURIComponent('${h.jsStringEscape(it.name)}')"/>
+            <f:textbox />
```
jglick, line 69: Can actually use
```suggestion
            <f:textbox/>
```

(The extra space is a relic of XHTML parsers IIUC.)

jglick (approved):
Nice and simple.

yaroslavafenkin (approved):
Cool approach with `@AncestorInPath`, applied in https://github.com/jenkinsci/matrix-project-plugin/pull/130/ too!

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
```diff
@@ -129,6 +132,11 @@ public List<FolderIconDescriptor> getIconDescriptors() {
         return r;
     }
 
+    public FormValidation doCheckDisplayNameOrNull(@AncestorInPath AbstractFolder folder, @QueryParameter String value) {
```
daniel-beck, line 135:
```suggestion
    @Restricted(NoExternalUse.class)
    public FormValidation doCheckDisplayNameOrNull(@AncestorInPath AbstractFolder folder, @QueryParameter String value) {
```

(Plus imports)

This plugin is a common dependency, and we don't want programmatic callers in other components for stuff like this.

Pldi23: Thank you Daniel, applied in [b4821ca](https://github.com/jenkinsci/cloudbees-folder-plugin/pull/268/commits/b4821ca080a49ba8a5874442445040b093386482)

daniel-beck (approved):

jglick (approved):

jtnord:
@jglick - how is this a `bug` what was broken before?  AFAICT this is an improvement?

jglick:
> what was broken before?

Just going by https://issues.jenkins.io/browse/JENKINS-60865

> Adding a content (using its hash) to the whitelist has the drawback to add weight on every response.

Open to interpretation. Really the distinction between `bug` and `enhancement` or even `developer` is fuzzy; my understanding is just that this change is something we want released.

PR #281, opened 2022-12-05
Update baseline, POM

jglick:
Supersedes #280.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,5 @@
 buildPlugin(useContainerAgent: true,
             configurations: [
-              [platform: 'linux', jdk: '11'],
+              [platform: 'linux', jdk: '17'],
               [platform: 'windows', jdk: '11'],
-              [platform: 'linux', jdk: '17']
 ])
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderComputation.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderComputation.java
@@ -72,6 +72,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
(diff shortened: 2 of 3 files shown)


Jenkinsfile
```diff
@@ -1,6 +1,5 @@
 buildPlugin(useContainerAgent: true,
             configurations: [
-              [platform: 'linux', jdk: '11'],
+              [platform: 'linux', jdk: '17'],
               [platform: 'windows', jdk: '11'],
-              [platform: 'linux', jdk: '17']
```
jglick, line 5: No need to waste time running Linux builds on both 11 and 17.

src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderComputation.java
```diff
@@ -122,7 +123,7 @@
     private volatile Result result;
 
     /** The past few durations for purposes of estimation. */
-    @CheckForNull
+    @Nullable // until set
```
jglick, line 126: Not sure why SpotBugs did not previously complain about https://github.com/jenkinsci/cloudbees-folder-plugin/blob/945ab2ef03dba53db977e7a47badd6d7c458224d/src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderComputation.java#L183-L189

jtnord: I do not see why this is not just initialised to an empty list here and made final simplifying the code elsewhere.
the null check here is only to prevent creation - and that will cause it to be set whenerver there is a run, and elsewhere the code is guarded by both the null check *and* an emptyness check.

Are we likely to have a Computation that will never run at least once?

jglick: > why this is not just initialised to an empty list here and made final

Because it would still be null in previously serialized computations.

jtnord: 🤦 for some reason I did not think this was serialized

PR #294, opened 2023-02-13
Replaced deprecated CaseInsensitiveComparator with java.lang.String one in tests

StefanSpieker:
Replaced deprecated CaseInsensitiveComparator with java.lang.String one in tests

* N/A

### Submitter checklist


jglick:
Is there a reason to change this here, when the occurrence is in old data (presumably some older version of Jenkins, if not the current, did save XML in this format)?

StefanSpieker:
I want to create a PR to remove the `CaseInsensitiveComparator` from `hudson.util` and I am not sure if there might be a unit test failure in a distant future because of this change.

jglick:
Well I think the existence of this test data is a good signal that you should _not_ remove that class, at least not without some careful analysis of where it might be used in `config.xml`s in the field, and a plan to migrate those.

jglick (changes requested):
Seems to be serving a useful function for checking backward compatibility of settings.

PR #332, opened 2023-07-17
chore: Parallelize Tests

rahulsom:
This will use all available cores to run tests, and parallelize by all - suite/class/method. On machines with multiple cores, this will vastly improve test performance. These times are on my M1 MBP with 10 (8P + 2E) cores. They were reported by maven on running `mvn verify`.

I first ran `mvn verify` and ignored the time.
Then I ran it thrice without this change, and thrice with this change.

All times in seconds.

|        | Run 1 | Run 2 | Run 3 | Average |
| ------ | ----: | ----: | ----: | ------: |
| Before |   183 |   189 |   188 |     187 |
| After  |    85 |    86 |    86 |      86 |
| Savings|       |       |       |     101 |
| %      |       |       |       |      54 |

This only changes the pom. No new tests are added. All existing tests pass.

* Internal: Parallelize Tests

### Submitter checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -107,4 +107,17 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <forkCount>1C</forkCount>
+          <parallel>all</parallel>
+          <useUnlimitedThreads>true</useUnlimitedThreads>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
 </project>


jtnord (changes requested):
this is already handled and hard codes the value which is not appropriate to all environments - as each surefire fork uses a specific amount of memory - a high CPU manchine without the same amount of RAM will have issues.

Thus tuning should be left for the environment where you are running.

e.g. just run `mvn -DforkCount=1C`  on your machine.

For CI this should be part of https://github.com/jenkins-infra/pipeline-library/tree/master/src that can be tuned per plugin.

rahulsom:
Hi @jtnord!

Unfortunately, not all tests can be parallelized. If a test relies on shared mutable state, or relies on mutations performed by a previous test, it's not going to run well in parallel. 
Some tests can be parallelized, but won't necessarily yield better outcomes. When the code under test is responsible for a forkjoin using a Threadpool, often that code sees no improvement because now we've got more Threadpools spinning up.

So I've been trying to identify plugins that do benefit from this and making the change to use all available cores.

Would it still be an improvement if I made this `0.5C`?

jglick:
> If a test relies on shared mutable state, or relies on mutations performed by a previous test,

…then that is a bug in the test which should be fixed. We expect anyone running Maven to be able to freely set a `forkCount` appropriate to their environment. (I do so locally via global profile.)

PR #347, opened 2023-10-03
Test with Java 21 and parent pom update

gounthar:
Test with Java 21
==

Java 21 was released Sep 19, 2023. We want to announce full support for Java 21 in early October and would like the most used plugins to be compiled and tested with Java 21.

The acceptance test harness and plugin bill of materials tests are already passing with Java 21. This is a further step to improve plugin readiness for use with Java 21 and for development with Java 21.

The change intentionally tests only two Java configurations, Java 17 and Java 21 because we believe that the risk of a regression that only affects Java 11 is shallow. We generate Java 11 byte code with the Java 17 and the Java 21 builds, so we're already testing Java 11 byte code.

Testing done
===

Confirmed tests pass with Java 21 after upgrading to a recent parent POM.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,5 +8,6 @@ buildPlugin(
   useContainerAgent: true, // Set to `false` if you need to use Docker for containerized tests
   configurations: [
     [platform: 'linux', jdk: 21],
+    [platform: 'linux', jdk: 11],
     [platform: 'windows', jdk: 17],
 ])
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,12 @@
-buildPlugin(useContainerAgent: true,
-            configurations: [
-              [platform: 'linux', jdk: '17'],
-              [platform: 'windows', jdk: '11'],
+/*
+ See the documentation for more options:
+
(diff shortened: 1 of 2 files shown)


alecharp (approved):
Looks good to me.

Jenkinsfile
```diff
@@ -1,5 +1,13 @@
...
+ See the documentation for more options:
+
+https://github.com/jenkins-infra/pipeline-library/
+
+*/
+buildPlugin(
+  useContainerAgent: true, // Set to `false` if you need to use Docker for containerized tests
+  configurations: [
+    [platform: 'linux', jdk: 21],
+    [platform: 'linux', jdk: 11],
```
jglick, line 11: Why the discrepancy from archetype?

alecharp: It's because of me. I didn't see that the archetype didn't require Java 11 testing anymore. I'll change that.

jtnord: it does not harm - some would say this is a good thing™️.

olamy: I tend to agree. If we say plugin xxxx still supports java 11, it's probably better to still have a build to test with it...

PR #348, opened 2023-10-03
Removes Java 11 testing

alecharp:
Following the archetype guidelines: https://github.com/jenkinsci/archetypes/blob/master/common-files/Jenkinsfile, the plugin should be tested on Linux with Java 21 and Windows on Java 17.

* Entry 1: Plugin code is built only on Java 17 and Java 21

### Submitter checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,6 +8,5 @@ buildPlugin(
   useContainerAgent: true, // Set to `false` if you need to use Docker for containerized tests
   configurations: [
     [platform: 'linux', jdk: 21],
-    [platform: 'linux', jdk: 11],
     [platform: 'windows', jdk: 17],
 ])


jtnord (changes requested):
please ping when we no longer support java 11 in the plugins baseline.

see #347

olamy:
> leaving java11 in the mix does no harm (infact it does good) and only impacts when the Jenkins baseline requires java 17 which is not for a while. This can therefore sit until we require the updated baseline for Jenkins.
> 
> the main aim of #347 was to test on Java 21 and that has been accomplished

Agree too. If I understand correctly the email on jdk baseline support. It was a jump from 11 to 21. So why having something such 17, 21 now whereas those plugin still have to support 11 and Jenkins core itself will jump from 11 to 21 (17 doesn't look to make sense)

alecharp:
I created this because I approved and merged https://github.com/jenkinsci/cloudbees-folder-plugin/pull/347 where I asked Bruno (outside of the pull request) to add the Java 11 into the build matrix when in fact the Java 11 build is not part of the plugin archetype anymore, and saw that with Jesse's comment.

> If I understand correctly the email on jdk baseline support. It was a jump from 11 to 21. So why having something such 17, 21 now whereas those plugin still have to support 11 and Jenkins core itself will jump from 11 to 21 (17 doesn't look to make sense)

In https://groups.google.com/g/jenkinsci-dev/c/RaAloTTM9CQ you mean? Jenkins running on Java 17 is supported until September 2025, and Jenkins running on Java 21 is supported starting September 2023 until March 2027. So there is no jump from 11 to 21. 

> leaving java11 in the mix does no harm (infact it does good)

To be fair, I'm not fully sure why building only with Java 17 and 21 wouldn't make a plugin not compatible with Java 11. 
If the problem is to keep the Java 11, then just close this pull request, it's ok.

olamy:
> I created this because I approved and merged #347 where I asked Bruno (outside of the pull request) to add the Java 11 into the build matrix when in fact the Java 11 build is not part of the plugin archetype anymore, and saw that with Jesse's comment.
> 
> > If I understand correctly the email on jdk baseline support. It was a jump from 11 to 21. So why having something such 17, 21 now whereas those plugin still have to support 11 and Jenkins core itself will jump from 11 to 21 (17 doesn't look to make sense)
> 
> In https://groups.google.com/g/jenkinsci-dev/c/RaAloTTM9CQ you mean? Jenkins running on Java 17 is supported until September 2025, and Jenkins running on Java 21 is supported starting September 2023 until March 2027. So there is no jump from 11 to 21.
> 

 reading here https://docs.google.com/spreadsheets/d/1Gc-0yuYOD5u674qnxbPOWhCU5t9viCJukVj_9b-kwAw/edit#gid=2094671884 I can see 11 still supported for long time.

Is core build only with 17 and 21 now? so why something different for plugins?

> > leaving java11 in the mix does no harm (infact it does good)
> 
> To be fair, I'm not fully sure why building only with Java 17 and 21 wouldn't make a plugin not compatible with Java 11. If the problem is to keep the Java 11, then just close this pull request, it's ok.

I only made a comment not blocking anything 
My point is: If we say 11 is supported we should build it/test it with 11 as well as newer (as core does)
it's just bad experience from the past....
but if the majority think the opposite and if this has been made mandatory somewhere else and approved by some democratic consensus that sounds like the right thing to do...

alecharp:
I was just trying to answer Jesse's concern, no big deal. Let's close this, I don't want to add more work or trouble to anyone.

jglick:
Please use https://github.com/jenkinsci/archetypes/issues/650 for general discussion.

basil:
As stated in https://github.com/jenkinsci/archetypes/issues/650 and related threads, the decision to drop Java 11 from the plugin test matrix early is a calculated risk (mitigated by continuing to maintain Java 11 builds in core, PCT, and ATH) with the benefit of reducing (by combining the "add Java 21" and "remove Java 11" steps into one step) the implementation cost of deploying these migrations throughout the `Jenkinsfile`s of critical plugins, especially in light of the fact that upstream (Eclipse Temurin) will drop support for Java 11 in about a year's time and these migrations tend to take months (if not years) to complete. Maintainers are free to express strong disagreement as long as they also do the work of keeping their `Jenkinsfile` up-to-date by adding Java 21 at the present time and removing Java 11 in a year's time.

PR #361, opened 2023-12-01
Miscellaneous code cleanup

basil:
Fixes various IDE warnings. Each change is a separate commit to ease reviewing. If any particular commit is unwanted, please just revert it and keep the others.

### Testing done

`mvn clean verify`


src/main/java/com/cloudbees/hudson/plugins/folder/computed/EventOutputStreams.java
```diff
@@ -171,7 +170,13 @@ private void offer(byte[] content) throws IOException {
                     // ignore
                 } finally {
                     appendNextOpen = true;
-                    IOUtils.closeQuietly(os);
+                    if (os != null) {
```
jglick, line 173: Simpler to just use a `try`-with-resources block.

jglick (approved):

PR #366, opened 2024-01-25
[JENKINS-72611] Forbid updating credentials ID

yaroslavafenkin:
See [JENKINS-72611](https://issues.jenkins.io/browse/JENKINS-72611).

* Forbid updating credentials ID

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
@@ -522,6 +522,14 @@ public class FolderCredentialsProvider extends CredentialsProvider {
         private synchronized boolean updateCredentials(@NonNull Domain domain, @NonNull Credentials current,
                                                        @NonNull Credentials replacement) throws IOException {
             checkPermission(CredentialsProvider.UPDATE);
+
+            // See IdCredentials.Helper#equals.
+            // Note: We probably should not assume that all credentials are IdCredentials. We also do this a few lines below in list.indexOf(current) call.
+            // If this one breaks, that one breaks as well.
+            if (!current.equals(replacement)) {
+                throw new IllegalArgumentException("Credentials' IDs do not match, will not update.");
+            }
+
             Map<Domain, List<Credentials>> domainCredentialsMap = getDomainCredentialsMap();
             if (domainCredentialsMap.containsKey(domain)) {
                 List<Credentials> list = domainCredentialsMap.get(domain);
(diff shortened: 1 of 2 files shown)


jglick:
complementing https://github.com/jenkinsci/credentials-plugin/pull/502

jglick (approved):

jtnord:
most likely needs to be reverted due to exactly the same reason as https://github.com/jenkinsci/credentials-plugin/pull/502

src/main/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProvider.java
```diff
@@ -522,6 +522,14 @@ private synchronized boolean removeCredentials(@NonNull Domain domain, @NonNull
         private synchronized boolean updateCredentials(@NonNull Domain domain, @NonNull Credentials current,
                                                        @NonNull Credentials replacement) throws IOException {
             checkPermission(CredentialsProvider.UPDATE);
+
+            // See IdCredentials.Helper#equals.
+            // Note: We probably should not assume that all credentials are IdCredentials. We also do this a few lines below in list.indexOf(current) call.
+            // If this one breaks, that one breaks as well.
+            if (!current.equals(replacement)) {
+                throw new IllegalArgumentException("Credentials' IDs do not match, will not update.");
```
jtnord, line 530: no no and more no.  you *must* handle credentials that are not IDCredentials

yaroslavafenkin: What about `int index = list.indexOf(current)` 6 lines below though? It will rely on `equals` under the hood, won't it?

jtnord: Yes it does.
But why do you think that matters?

yaroslavafenkin: If credential isn't an `IdCredential`, would that `indexOf` correctly find a credential to be updated?

jtnord: > If credential isn't an IdCredential, would that indexOf correctly find a credential to be updated?

Under what conditions would it fail to find `current` in a list of `Credentials`?

yaroslavafenkin: `Object.equals`.
I've tested locally with the loadfocus plugin, and updating indeed works there. I'm surprised that object reference is that same. I don't know why I assumed it wouldn't be.

PR #367, opened 2024-01-26
[JENKINS-72613] `ItemGroup.getItemName` override

jglick:
Downstream of https://github.com/jenkinsci/jenkins/pull/8902, amending https://github.com/jenkinsci/cloudbees-folder-plugin/pull/82.

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
@@ -497,6 +497,33 @@ public abstract class AbstractFolder<I extends TopLevelItem> extends AbstractIte
         return configurations;
     }
 
+    @Override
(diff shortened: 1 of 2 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -497,6 +497,33 @@ public static <K, V extends TopLevelItem> Map<K, V> loadChildren(AbstractFolder<
         return configurations;
     }
 
+    @Override
```
jglick, line 500: Note: actually required, or the `interface default` implementation will be used instead.

basil (commented):
Is this not copypasta of existing code in this file?

basil:
Suggest refactoring the existing code to avoid duplication as a prerequisite to this PR to avoid increasing technical debt.

PR #372, opened 2024-02-09
[JENKINS-72613] ItemGroup.getItemName override & refactor

Vlatombe:
A more invasive approach to #367, including refactor of `AbstractFolder` & `ChildNameGenerator` to avoid the copy-pasta lookalike code that was caused by the previous design. 

* ...

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
@@ -497,6 +497,33 @@ public abstract class AbstractFolder<I extends TopLevelItem> extends AbstractIte
         return configurations;
     }
 
+    @Override
+    public String getItemName(File dir, I item) {
+        ChildNameGenerator<AbstractFolder<I>, I> childNameGenerator = childNameGenerator();
+        String itemName;
+        if (childNameGenerator == null) {
+            itemName = dir.getName();
+        } else {
+            File nameFile = new File(dir, ChildNameGenerator.CHILD_NAME_FILE);
+            if (nameFile.isFile()) {
+                try {
+                    itemName = StringUtils.trimToNull(Files.readString(nameFile.toPath(), StandardCharsets.UTF_8));
+                } catch (IOException e) {
+                    LOGGER.log(Level.WARNING, e, () ->"Caught an IOException while trying to read " + nameFile + ", assuming child name is " + dir.getName());
+                    itemName = null;
+                }
+                if (itemName == null) {
+                    LOGGER.log(Level.WARNING, "{0} was empty, assuming child name is {1}", new Object[]{nameFile, dir.getName()});
+                    itemName = dir.getName();
+                }
+            } else {
+                itemName = dir.getName();
+            }
+        }
+        String finalItemName = itemName;
+        LOGGER.log(Level.FINE, () -> "itemName = " + finalItemName);
+        return itemName;
+    }
 
     protected final I itemsPut(String name, I item) {
         ChildNameGenerator<AbstractFolder<I>, I> childNameGenerator = childNameGenerator();
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
@@ -47,7 +47,6 @@ import hudson.model.Failure;
 import hudson.model.HealthReport;
 import hudson.model.Item;
 import hudson.model.ItemGroup;
-import hudson.model.ItemGroupMixIn;
 import hudson.model.Items;
 import hudson.model.Job;
 import hudson.model.ModifiableViewGroup;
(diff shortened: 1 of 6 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -80,6 +88,7 @@
  */
 // TODO migrate this functionality (by changing the base class) into core once baseline Jenkins has JENKINS-41222 merged
```
jglick, line 89: Perhaps obsolete.

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -185,6 +194,73 @@ protected final String idealNameFromItem(@NonNull P parent, @NonNull I item) {
     @CheckForNull
     public abstract String dirNameFromItem(@NonNull P parent, @NonNull I item);
 
+    public static class ResultWithOptionalSave<I> {
```
jglick, line 197: Can this and `ensureItemDirectory` be restricted to internal use, or package-scope? (`ChildNameGenerator` as a whole is used by `branch-api`.)

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
```diff
@@ -216,9 +216,41 @@ public boolean isLookAndFeelConfigurable(AbstractFolder<?> folder) {
      */
     // TODO figure out how one could un-wind name mangling if one ever wanted to
     // TODO move the name mangling code from branch-api to this plugin so that everyone can use it
-    @CheckForNull
+    @NonNull
     public <I extends TopLevelItem> ChildNameGenerator<AbstractFolder<I>,I> childNameGenerator() {
-        return null;
+        return new LegacyChildNameGenerator<>();
```
jglick, line 221: Overridden by `MultiBranchProjectDescriptor` and `OrganizationFolder.DescriptorImpl`, OK.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -337,189 +334,73 @@ public static <K, V extends TopLevelItem> Map<K, V> loadChildren(AbstractFolder<
...
-            } else {
-                for (V item : parent.items.values()) {
-                    String itemName = childNameGenerator.dirNameFromItem(parent, item);
-                    if (itemName == null) {
-                        itemName = childNameGenerator.dirNameFromLegacy(parent, item.getName());
-                    }
-                    byDirName.put(itemName, item);
-                }
+            for (V item : parent.items.values()) {
+                byDirName.put(childNameGenerator.dirNameFromItem(parent, item), item);
```
jglick, line 338: `dirNameFromItem` is declared to return null (though the reason is undocumented), and actual impls in `branch-api` do so at least in some (perhaps rare) conditions, so you still need to check for that I suppose. Maybe have a wrapper that uses `LegacyChildNameGenerator` as a fallback in case the real impl returns null from `itemNameFromItem` or `dirNameFromItem`?

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -614,11 +491,8 @@ protected final File getRootDirFor(String name) {
 
     @Override
     public File getRootDirFor(I child) {
-        ChildNameGenerator<AbstractFolder<I>,I> childNameGenerator = childNameGenerator();
-        if (childNameGenerator == null) {
-            return getRootDirFor(child.getName());
-        }
-        String name = childNameGenerator.dirNameFromItem(this, child);
+        var childNameGenerator = childNameGenerator();
+        var name = childNameGenerator.dirNameFromItem(this, child);
```
jglick, line 495: Switching to `var` here makes the diff needlessly long.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolderDescriptor.java
```diff
@@ -216,9 +216,41 @@ public boolean isLookAndFeelConfigurable(AbstractFolder<?> folder) {
      */
     // TODO figure out how one could un-wind name mangling if one ever wanted to
```
jglick, line 217: Not sure what this meant, but probably long obsolete?

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -185,6 +194,73 @@ protected final String idealNameFromItem(@NonNull P parent, @NonNull I item) {
...
+    }
+
+    /**
+     * Ensures that the item is stored in the correct directory, and moves it if necessary.
+     *
+     * @param parent the parent of the given item.
+     * @param item the item to determine directory for.
+     * @param legacyDir The directory name that we are loading an item from.
+     * @return a reference to the (new) directory storing the item, and whether the item needs to be saved.
+     * @throws IOException
```
jglick, line 230:
```suggestion
```

(or actually document)

jglick (commented):
Other than what looks like a missing null check, seems OK I think? I never fully understood the original code so I cannot comment deeply.

Be sure to test against at least `branch-api` and https://github.com/jenkinsci/workflow-multibranch-plugin/pull/292.

jglick (approved):

Vlatombe:
> Be sure to test against at least branch-api and https://github.com/jenkinsci/workflow-multibranch-plugin/pull/292.

Both green

PR #376, opened 2024-02-14
[JENKINS-72613] ItemGroup.getItemName override & refactor, take 2

Vlatombe:
This is a re-file of #372 with extra fixes based on PCT test failures observed internally. Should go through a BOM build for extra sanity testing.

* ...

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
@@ -47,7 +47,6 @@ import hudson.model.Failure;
 import hudson.model.HealthReport;
 import hudson.model.Item;
 import hudson.model.ItemGroup;
-import hudson.model.ItemGroupMixIn;
 import hudson.model.Items;
 import hudson.model.Job;
 import hudson.model.ModifiableViewGroup;
@@ -71,9 +70,8 @@ import hudson.util.HttpResponses;
 import hudson.views.DefaultViewsTabBar;
 import hudson.views.ViewsTabBar;
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
-import java.nio.charset.StandardCharsets;
-import java.nio.file.Files;
 import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -102,7 +100,6 @@ import jenkins.model.ProjectNamingStrategy;
 import jenkins.model.TransientActionFactory;
 import net.sf.json.JSONObject;
 import org.acegisecurity.AccessDeniedException;
-import org.apache.commons.lang.StringUtils;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.HttpRedirect;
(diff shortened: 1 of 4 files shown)


jglick (approved):
OK assuming `bom` passes.

Vlatombe:
https://github.com/jenkinsci/bom/pull/2927

Vlatombe:
All green https://github.com/jenkinsci/bom/pull/2927/checks

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -337,216 +334,73 @@ public static <K, V extends TopLevelItem> Map<K, V> loadChildren(AbstractFolder<
...
         }
         for (File subdir : subdirs) {
+            File effectiveSubdir = subdir;
+            String childName = childNameGenerator.readItemName(subdir);
+            // Try to retain the identity of an existing child object if we can.
+            V itemFromDir;
+            V item;
+            boolean itemNeedsSave = false;
+            String name;
+            item = itemFromDir = byDirName.get(childName);
```
basil, line 349: @vlatombe Note that the keys of `byDirName` have gone through `NameMangler` if they are an `OrganizationFolder` or `MultiBranchProjectDescriptor` with special characters. Before this PR, `childName` was reset to `subdir.getName()` after reading the child name file, so the lookup was correct. After this PR, the lookup is done with the unmangled name rather than the `NameMangler`-mangled name. Therefore, the attempt to "try to retain the identity of an existing child object if we can" fails. This is a regression.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -337,216 +334,73 @@ public static <K, V extends TopLevelItem> Map<K, V> loadChildren(AbstractFolder<
...
-            if (nameFile.isFile()) {
-                try {
-                    oldName = StringUtils.trimToNull(Files.readString(nameFile.toPath(), StandardCharsets.UTF_8));
-                } catch (IOException e) {
-                    oldName = null;
-                }
-            } else {
-                oldName = null;
-            }
-            if (!name.equals(oldName)) {
```
basil, line 542: @Vlatombe Before this PR, we only attempted to write the child name file when the new name was not equal to the old name. After this PR, we started unconditionally writing the child name file, which at the very minimum introduced a performance regression (updating the file modification time unnecessarily) and may have introduced a worse regression if other consumers tried to read the file while it was being (unnecessarily) modified.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
Vlatombe: #486

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
Vlatombe: Re-reading the code, the logic just moved to https://github.com/jenkinsci/cloudbees-folder-plugin/blob/5338bc4d6992a1b0791360f7793fbb650aeffaca/src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java#L327-L339. I think the behaviour is the same as previously.

basil: @vlatombe Only after I fixed the regression in https://github.com/jenkinsci/cloudbees-folder-plugin/pull/482.

Vlatombe: Ah, right. Thank you for it!

PR #395, opened 2024-04-30
[JENKINS-73086] Allow users with Overall/Manage permission to configure global default folder settings

mikecirioli:
See [JENKINS-73086](https://issues.jenkins-ci.org/browse/JENKINS-73086).

* ...

### Submitter checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,9 @@
     <jenkins.version>2.454</jenkins.version>
     <no-test-jar>false</no-test-jar>
     <hpi.compatibleSinceVersion>5.2</hpi.compatibleSinceVersion>
+    <useBeta>true</useBeta> <!-- needed for Jenkins.MANAGE support -->
   </properties>
+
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -24,7 +24,9 @@
     <jenkins.version>2.454</jenkins.version>
     <no-test-jar>false</no-test-jar>
     <hpi.compatibleSinceVersion>5.2</hpi.compatibleSinceVersion>
+    <useBeta>true</useBeta> <!-- needed for Jenkins.MANAGE support -->
```
jtnord, line 27: are there no Plans / PRs to remove the @Beta from [the permission](https://github.com/jenkinsci/jenkins/blob/33858d1c673efe415b2daabb9c54dc04a05b314d/core/src/main/java/jenkins/model/Jenkins.java#L5769-L5774)?

jtnord (approved):

PR #419, opened 2024-08-22
Remove disable button to match other job type since 2.460

jonesbusy:
Remove disable button to match other job type since 2.460

I can open an Jira ticket if needed

* Remove disable button to match other job type since 2.460

### Submitter checklist

### DIFF
--- a/src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
+++ b/src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
@@ -609,45 +609,6 @@ public class ComputedFolderTest {
         waitUntilNoActivityIgnoringThreadDeathUpTo(10000);
         d.checkRename("d2");
     }
-    
-    @Issue("JENKINS-60900")
-    @Test
-    public void enabledAndDisableFromUi() throws Exception {
-        SampleComputedFolder folder = r.jenkins.createProject(SampleComputedFolder.class, "d");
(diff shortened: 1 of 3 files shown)


src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
```diff
@@ -609,45 +609,6 @@ public void renameWhileComputing() throws Exception {
         waitUntilNoActivityIgnoringThreadDeathUpTo(10000);
         d.checkRename("d2");
     }
-    
-    @Issue("JENKINS-60900")
-    @Test
-    public void enabledAndDisableFromUi() throws Exception {
```
jglick, line 615: #143

src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
```diff
@@ -609,45 +609,6 @@ public void renameWhileComputing() throws Exception {
...
-        cfg = (HtmlForm)r.createWebClient().getPage(folder).getElementById("enable-project");
-        assertNotNull(cfg);
-        // Re enable the folder
-        r.submit(cfg);
-        assertFalse(folder.isDisabled());
-    }
-
-    @Issue("JENKINS-66168")
-    @Test
-    public void enabledAndDisableFromUiViews() throws Exception {
```
jglick, line 632: #185

jglick (approved):

PR #424, opened 2024-09-11
Folder view matches look of job view

strangelookingnerd:
Comparing the view of a folder vs. job I noticed some minor differences this PR aims to fix by adapting the definitions in `AbstractFolder/view-index-top.jelly` to the [core](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/hudson/model/Job/index.jelly).

Most notable changes:
* same size of text and icon
* same display of the "full name" in case a folder is nested

See the following screenshots for comparissions:

### Before
| Job | Folder |
|-|-|
| grafik | grafik | 
| grafik | grafik |

### After
| Job | Folder |
|-|-|
| grafik | grafik |
| grafik | grafik |

One thing that gave me a hard time is the `description` - there seem to be multiple of those for the folder and the views? Is that intented? I will need some clarification on this before merging to not introduce regressions.

* Entry 1: Folder view matches look of job view

### Submitter checklist

### DIFF
--- a/src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/view-index-top.jelly
+++ b/src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/view-index-top.jelly
@@ -42,9 +42,6 @@ THE SOFTWARE.
         <l:breakable value="${it.displayName}"/>
       </h1>
     </div>
-    <div class="jenkins-app-bar__controls">
-      <t:editDescriptionButton description="${it.primaryView.description}" permission="${it.CONFIGURE}"/>
-    </div>
   </div>
   <j:if test="${(it.fullName!=it.fullDisplayName)}">
     <j:choose>
@@ -57,11 +54,12 @@ THE SOFTWARE.
     </j:choose>
   </j:if>
   <div id="view-message">
-    <div id="systemmessage">
-      <j:out value="${app.systemMessage!=null ? app.markupFormatter.translate(app.systemMessage) : ''}"/>
-    </div>
-    <t:editableDescription description="${it.primaryView.description}" permission="${it.CONFIGURE}" hideButton="true"/>
+      <div id="systemmessage">
+          <j:out value="${app.systemMessage!=null ? app.markupFormatter.translate(app.systemMessage) : ''}"/>
+      </div>
+      <j:out value="${it.description!=null ? app.markupFormatter.translate(it.description) : ''}"/>
   </div>
+  <this:description />
   <j:choose>
       <j:when test="${!it.supportsMakeDisabled()}">
           <!-- for now, quietly omit the option -->
--- a/src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/view-index-top.jelly
+++ b/src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/view-index-top.jelly
@@ -24,38 +24,46 @@ THE SOFTWARE.
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout"
          xmlns:t="/lib/hudson" xmlns:f="/lib/form" xmlns:i="jelly:fmt" xmlns:this="this">
-  <j:set var="iconClass" value="${it.iconColor.getIconClassName()}"/>
-  <j:if test="${iconClass != null}">
-      <j:set var="iconClass" value="${iconClass} icon-xlg"/>
-  </j:if>
-  <h1>
+  <div class="jenkins-app-bar">
+    <div class="jenkins-app-bar__content jenkins-build-caption">
+      <j:set var="iconClass" value="${it.iconColor.getIconClassName()}"/>
+      <j:if test="${iconClass != null}">
+        <j:set var="iconClass" value="${iconClass} icon-lg"/>
+      </j:if>
+      <j:choose>
+        <j:when test="${iconClass != null}">
+          <l:icon class="${iconClass}" alt="${it.iconColor.description}" tooltip="${it.iconColor.description}" />
+        </j:when>
+        <j:otherwise>
+          <l:icon src="${it.iconColor.getImageOf('32x32')}" alt="${it.iconColor.description}" tooltip="${it.iconColor.description}" />
+        </j:otherwise>
+      </j:choose>
+      <h1 class="job-index-headline page-headline">
+        <l:breakable value="${it.displayName}"/>
+      </h1>
+    </div>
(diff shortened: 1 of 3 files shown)


strangelookingnerd:
Maybe someone from @jenkinsci/sig-ux could also provide feedback

timja:
The ideal would be to reuse core and use that instead so that changes automatically get adapted.

If its possible to adapt here / core / both that would be the best solution.

If its not then what you've done here looks good UX wise (I've not tested it manually).

The main changes I saw were:
* moving description control to the app bar
* smaller icon
* full project name being added

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/view-index-top.jelly
```diff
@@ -24,36 +24,39 @@ THE SOFTWARE.
...
-  </h1>
-  <j:if test="${it.name != it.displayName}">
-    ${%Folder name}: ${it.fullName}
   </j:if>
-  <div id="view-message">
-      <div id="systemmessage">
-          <j:out value="${app.systemMessage!=null ? app.markupFormatter.translate(app.systemMessage) : ''}"/>
-      </div>
-      <j:out value="${it.description!=null ? app.markupFormatter.translate(it.description) : ''}"/>
-  </div>
```
jtnord, line 50: this removes the system message from the folder.

## before

image
image


## after
image
image

jtnord (changes requested):
this has removed the system message from folders.  (not tested anything else)

it has also gone from core and it appears this was not noticed - will file a ticket for that.

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/view-index-top.jelly
strangelookingnerd: It is not shown for Jobs, so I'm wondering, do we actually _want_ the system message to be displayed there?

jtnord: > It is not shown for Jobs, so I'm wondering, do we actually _want_ the system message to be displayed there?

I would argue we do want it for `Items` (and more..) (and it was never implemented¹).  
if an admin uses a system message to convey some "downtime" then users interacting with Jenkins would never see it unless they where on the root page (which is unlikely if they follow links from say GH checks or just go to thier job).

¹ I actually have a (likely false) memory that it did used to show up for Items also!

jtnord: I would suggest that the discussion about where the systemMessage should or should not be should happen outside this PR as it impacts usability over and above the other changes.

jtnord:
> One thing that gave me a hard time is the description - there seem to be multiple of those for the folder and the views? Is that intented?

You mean a folder has a description and a view can also have one, so when using a view for the folder you have potentially 2 conflicting descriptions?
This is the case for Jenkins core (where jenkins has getDescription() returning the system message yet the view for it also has a description.  In this case both are shown.
if that is correct / desired behavour I am not sure.  Could probably be argued either way (and possibly depends on the view in use), but one thing is for sure, without a request to change it someone will be relying on the existing behaviour and a change without justification will be a terrible regression for them :)

strangelookingnerd:
> > One thing that gave me a hard time is the description - there seem to be multiple of those for the folder and the views? Is that intented?
> 
> You mean a folder has a description and a view can also have one, so when using a view for the folder you have potentially 2 conflicting descriptions? This is the case for Jenkins core (where jenkins has getDescription() returning the system message yet the view for it also has a description. In this case both are shown. if that is correct / desired behavour I am not sure. Could probably be argued either way (and possibly depends on the view in use), but one thing is for sure, without a request to change it someone will be relying on the existing behaviour and a change without justification will be a terrible regression for them :)

As I understand it the `AbstractFolder` has a `description` field inherited from `AbstractItem` - this field can be configured but is only visible on the configuration page and nowhere else.
Then the `AbstractFolder` has a `View` that has its own `description` field which is actually shown in the overview and can only be edited there - but not on the configuration page.

I agree, introducing regressions is not my intent, therefor I left it as-is.

jtnord:
> > > One thing that gave me a hard time is the description - there seem to be multiple of those for the folder and the views? Is that intented?
> > 
> > 
> > You mean a folder has a description and a view can also have one, so when using a view for the folder you have potentially 2 conflicting descriptions? This is the case for Jenkins core (where jenkins has getDescription() returning the system message yet the view for it also has a description. In this case both are shown. if that is correct / desired behavour I am not sure. Could probably be argued either way (and possibly depends on the view in use), but one thing is for sure, without a request to change it someone will be relying on the existing behaviour and a change without justification will be a terrible regression for them :)
> 
> As I understand it the `AbstractFolder` has a `description` field inherited from `AbstractItem` - this field can be configured but is only visible on the configuration page and nowhere else. 

I see it in the folder page.

> Then the `AbstractFolder` has a `View` that has its own `description` field which is actually shown in the overview and can only be edited there - but not on the configuration page.

👍 

> I agree, introducing regressions is not my intent, therefor I left it as-is.

from `master`

image

strangelookingnerd:
> from `master` I see both descriptions when set

Well, now that's embarassing - sorry for the confusion 🤭 
I'll try and fix it.

strangelookingnerd:
Restored the original description beavior in https://github.com/jenkinsci/cloudbees-folder-plugin/pull/424/commits/1a781baaba8e7f259dd2d26c0acfa0448df4ccf3 - edited the PR description.

jtnord (dismissed):

jtnord:
@strangelookingnerd is this complete from your point of view before i merge?

strangelookingnerd:
> @strangelookingnerd is this complete from your point of view before i merge?

Yes, thanks for taking care of it 👍🏼

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/view-index-top_ja.properties
```diff
@@ -21,3 +21,4 @@
 # THE SOFTWARE.
 
 Folder\ name=\u30d5\u30a9\u30eb\u30c0\u540d
+Full\ folder\ name=\u5b8c\u5168\u306a\u30d5\u30a9\u30eb\u30c0\u540d
```
jglick, line 24: Uh, do you speak Japanese fluently? As a general rule, no one other than translators should touch localized files except perhaps to delete keys deleted from the base file, or move keys moved in base files.

janfaracik (approved):
Tried it, looks good to me. I'd be keen to see some of these changes in `core` eventually, e.g. the ability to show an icon in the app bar of a view.

PR #428, opened 2024-09-25
Gratuitous `save` from `AbstractFolder.onDeleted/onRenamed`

jglick:
While testing some scenarios involving job deletion in CloudBees CI, we found some anomalous events caused by a save event on `config.xml` of the parent folder. This makes no apparent sense, because the folder is not being modified.


dwnusbaum (approved):

jglick:
the [root analogue](https://github.com/jenkinsci/jenkins/blob/435bb7989126d6ef3e3d78ddc5597354b71bff74/core/src/main/java/jenkins/model/Jenkins.java#L3287-L3295), for the record

Vlatombe (approved):

PR #433, opened 2024-10-16
[JENKINS-73930] Put back the disabled warning

Dohbedoh:
See [JENKINS-73930](https://issues.jenkins.io/browse/JENKINS-73930). Amend 

The button was removed from https://github.com/jenkinsci/cloudbees-folder-plugin/pull/419. But along with the warning. It is not obvious anymore when a folder is disabled.

Before:

before

After:

after

* Re-instaore the Disabled warning when a Folder is disabled

### Submitter checklist

### DIFF
--- a/src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
+++ b/src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
@@ -630,22 +630,14 @@ public class ComputedFolderTest {
     public void disabledWarningFromUiViews() throws Exception {
         LockedDownSampleComputedFolder folder = r.jenkins.createProject(LockedDownSampleComputedFolder.class, "d");
         assertFalse("by default, a folder is disabled", folder.isDisabled());
-        folder.getViews().forEach(view -> {
-            try {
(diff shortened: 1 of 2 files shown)


jonesbusy (approved):

jonesbusy (approved):

src/test/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolderTest.java
```diff
@@ -625,6 +625,29 @@ public void failAllDeletedOnes() throws Exception {
...
+    @Issue("JENKINS-73930")
+    @Test
+    public void disabledWarningFromUiViews() throws Exception {
+        LockedDownSampleComputedFolder folder = r.jenkins.createProject(LockedDownSampleComputedFolder.class, "d");
+        assertFalse("by default, a folder is disabled", folder.isDisabled());
+        folder.getViews().forEach(view -> {
+            try {
+                assertNull(r.createWebClient().goTo(view.getViewUrl()).getElementById("disabled-message"));
+            } catch (Exception e) {
+                Assert.fail();
```
jglick, line 637: Better to switch to a plain `for` loop so you do not need to worry about checked exceptions.

jglick (approved):

PR #441, opened 2024-11-14
[JENKINS-72958] expose buildable property to API

jtnord:
https://github.com/jenkinsci/branch-api-plugin/pull/468 but on the base class that defines `isBuildable`

See [JENKINS-73425](https://issues.jenkins.io/browse/JENKINS-73425).

Obsoletes https://github.com/jenkinsci/branch-api-plugin/pull/468 (untested)

* Expose `buildable` for ComputedFolders to the REST API

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
@@ -84,6 +84,7 @@ import org.kohsuke.stapler.HttpResponses;
 import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerRequest;
 import org.kohsuke.stapler.StaplerResponse;
+import org.kohsuke.stapler.export.Exported;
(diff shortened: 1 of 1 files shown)


jglick (approved):

jglick:
(pending testing as per https://github.com/jenkinsci/branch-api-plugin/pull/468#issuecomment-2476378328)

rsandell (approved):

MarkRx:
@jtnord tested works thanks!

http://localhost:8080/jenkins/job/folder1/job/pipeline1/api/json?tree=buildable

Enabled:
```
{
    _class: "org.jenkinsci.plugins.workflow.multibranch.WorkflowMultiBranchProject",
    buildable: true
}
```

Disabled:
```
{
    _class: "org.jenkinsci.plugins.workflow.multibranch.WorkflowMultiBranchProject",
    buildable: false
}
```

Note child jobs will have buildable=true even if the parent is disabled (buildable=false). The buildable property doesn't propagate so the parent job needs to be checked. This is existing behavior and unrelated to exposing the buildable property.

PR #444, opened 2024-11-18
Migrate from EE 8 to EE 9

basil:
See jenkinsci/jenkins#9672.

### Testing done

- Passing CI build
- Tested in context in `branch-api` and `matrix-auth` before and after migrating those plugins to EE 9
- ATH (weekly and LTS): passing
- BOM (weekly): passing

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
@@ -70,6 +70,9 @@ import hudson.util.FormValidation;
 import hudson.util.HttpResponses;
 import hudson.views.DefaultViewsTabBar;
 import hudson.views.ViewsTabBar;
+import io.jenkins.servlet.ServletExceptionWrapper;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.http.HttpServletResponse;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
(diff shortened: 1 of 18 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -1092,13 +1159,39 @@ public void renameTo(String newName) throws IOException {
      * {@inheritDoc}
      */
     @Override
-    public synchronized void doSubmitDescription(StaplerRequest req, StaplerResponse rsp) throws IOException, ServletException {
+    public synchronized void doSubmitDescription(StaplerRequest2 req, StaplerResponse2 rsp) throws IOException, ServletException {
+        if (Util.isOverridden(AbstractFolder.class, getClass(), "doSubmitDescription", StaplerRequest.class, StaplerResponse.class)) {
```
jglick, line 1163: Most if not all of these web methods could probably be simplified considerably by just deleting the deprecated overload, which I doubt is either called programmatically or overridden by any known subclass. Pending that, it would make sense to mark the new overloads `final` and/or `@Restricted(NoExternalUse.class)` unless and until some legitimate use case for overriding one is identified.

(not referring to `submit`, which _is_ clearly overridden at least within this plugin, and also in `branch-api`)

basil: Does not seem necessary for this PR.

alecharp (approved):
This looks complete to me for the EE 8 to 9 migration. 
Additional changes wouldn't be related to the migration and would require additional research as to their usages externally to the plugin.

jglick:
> Additional changes wouldn't be related to the migration

Well, no IMO, the PR is introducing new externally-visible API signatures which are not expected to ever be used externally, so it is related. Not blocking.

basil:
But they are only being introduced where there were existing externally-visible API signatures for the old EE 8 types. In other words, the fact that methods were exposed but unused before this PR was just technical debt, which I am not making worse in this PR (but also not making any better).

PR #447, opened 2024-12-03
Defend against null `AbstractFolder.primaryView`

jglick:
Not observed as such, but seems possible from code inspection. See https://github.com/jenkinsci/jenkins/pull/10023.


Vlatombe (approved):

dwnusbaum (approved):
FWIW, in practice I have seen this issue prior to https://github.com/jenkinsci/hashicorp-vault-plugin/pull/336, caused by earlier mistakes in that plugin, although in that case I think https://github.com/jenkinsci/jenkins/pull/9653 would have prevented the issue.

PR #452, opened 2024-12-15
Add icons for search

timja:
~Requires https://github.com/jenkinsci/branch-api-plugin/pull/497 to be taken out of draft~ - bit of a hack here instead due to a subclass referencing this class and couldn't see an easy fix

Part of https://github.com/jenkinsci/jenkins/pull/10049 but can be merged interdependently 

image

* ...

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
@@ -101,6 +101,7 @@ import jenkins.model.ModelObjectWithChildren;
 import jenkins.model.ProjectNamingStrategy;
 import jenkins.model.TransientActionFactory;
 import net.sf.json.JSONObject;
+import org.jenkins.ui.icon.IconSpec;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.Beta;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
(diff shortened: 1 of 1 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -305,6 +306,15 @@ protected AbstractFolderViewHolder newFolderViewHolder() {
         return new DefaultFolderViewHolder(views, null, newDefaultViewsTabBar());
     }
 
+    @Override
+    public String getIconClassName() {
+        // avoid https://issues.jenkins.io/browse/JENKINS-74990
+        if (icon.getClass().getName().equals("jenkins.branch.MetadataActionFolderIcon")) {
```
timja, line 312: not ideal but I couldn't see a cleaner solution.

This if block means that the `MetadataActionFolderIcon` falls back into the code it previously did before `Folder` had an `IconSpec`.

in command palette it works fine because e.g. `OrganizationFolder` overrides `getIconClassName` so it gets the right icon, in the example I tried (GitHub organisation) it didn't load the icon in command palette but that's because it was using the `IconSet` API and not symbols which isn't supported by command palette right now (cc @janfaracik )

timja:
@jenkinsci/cloudbees-folder-plugin-developers  ?

jglick:
> Requires https://github.com/jenkinsci/branch-api-plugin/pull/497 to be taken out of draft

That is closed; was this part of the description outdated?

jglick:
(merge conflicts here BTW)

timja:
> > Requires [jenkinsci/branch-api-plugin#497](https://github.com/jenkinsci/branch-api-plugin/pull/497) to be taken out of draft
> 
> That is closed; was this part of the description outdated?

outdated yes, updated now

jglick (approved):
No clue what this is doing but sounds like it fixes something.

PR #468, opened 2025-02-11
Support groups for Command Palette

janfaracik:
Adds support for grouping via https://github.com/jenkinsci/jenkins/pull/10252. Previously folders would appear under the 'Other' group, rather than 'Items'.

**Before**
image

**After**
image

* Support groups for Command Palette

### Submitter checklist


pom.xml
```diff
@@ -23,7 +23,7 @@
     <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
     <!-- https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/ -->
     <jenkins.baseline>2.479</jenkins.baseline>
-    <jenkins.version>${jenkins.baseline}.1</jenkins.version>
```
janfaracik, line 26: Need to change to latest weekly or LTS when groups are released.

PR #472, opened 2025-03-01
apply jenkins styling to selects

mawinter69:
also remove the obsolete blockwrapper

Before:
Selecting the default view
image

Moving an item
image

After:
Selecting the default view: 
image

Moving an item
image

### Testing done:
Gone to move page of a job and verified that it still works

* use jenkins styling for  selects

### Submitter checklist


src/main/resources/com/cloudbees/hudson/plugins/folder/computed/FolderComputation/index.jelly
```diff
@@ -34,7 +33,7 @@ xmlns:fo="/lib/folder">
                     (<i:formatDate value="${it.timestamp.time}" type="both" dateStyle="medium" timeStyle="medium"/>)
                 </j:if>
             </t:buildCaption>
-            <fo:blockWrapper>
+            <table style="margin-top: 1em; margin-left:1em;">
```
mawinter69, line 36: Actually it was wrong when the blockwrapper was introduced to use it here. The `t:summary` is explicitly requiring a surrounding table as it itself creates a `<tr>`
Though I couldn't find a way how to make this page show up anywhere. I think it would require having a ComputedFolder that keeps kind of an execution history of the computations but branch-api doesn't do that

alecharp:
Might be a lot to ask, but could you introduce before/after screenshots in your pull request description? Thanks @mawinter69

mawinter69:
> Might be a lot to ask, but could you introduce before/after screenshots in your pull request description? Thanks @mawinter69

done

mawinter69:
the test error looks unrelated

src/main/resources/com/cloudbees/hudson/plugins/folder/config/AbstractFolderConfiguration/config.jelly
```diff
@@ -3,13 +3,7 @@
     <f:section title="${%Folder}">
         <f:entry title="${%Health Metrics}"
                  description="${%List of default health metrics for Folders}">
-            <f:repeatableHeteroProperty field="healthMetrics" oneEach="true" honorOrder="true" hasHeader="true">
-                <f:entry title="">
-                    <div align="right">
-                        <f:repeatableDeleteButton/>
-                    </div>
-                </f:entry>
```
jglick, line 11: Huh, when did this become unnecessary?

timja: I went back 18 years and it was there then: https://github.com/jenkinsci/jenkins/blob/dddb4598df25527220a1e0f46d1a0d25217dd4c8/core/src/main/resources/lib/form/hetero-list.jelly#L95

PR #479, opened 2025-04-04
Check to see if `idealNames` workaround can be removed

basil:
This workaround is very old, so checking to see if it is still needed.

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
@@ -27,20 +27,15 @@ package com.cloudbees.hudson.plugins.folder;
 import com.cloudbees.hudson.plugins.folder.computed.ComputedFolder;
 import hudson.BulkChange;
 import hudson.Util;
-import hudson.model.AbstractItem;
 import hudson.model.Action;
 import hudson.model.Item;
-import hudson.model.ItemGroup;
 import hudson.model.JobProperty;
 import hudson.model.TopLevelItem;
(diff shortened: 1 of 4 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -99,58 +85,6 @@ public abstract class ChildNameGenerator<P extends AbstractFolder<I>, I extends
...
-     * @param project the {@link AbstractFolder}.
-     * @param itemName the name that will be returned by {@link Item#getName()} when the item is constructed. This is
-     *                 the second parameter of {@link AbstractItem#AbstractItem(ItemGroup, String)}. This one would be
-     *                 the one with URL path segment escaping.
-     * @param idealName the original name before whatever URL path segment escaping you applied
-     * @return the {@link Trace} to keep track of when we can remove the memory of the creation process. Please
-     * {@link Trace#close()} the trace after the item is created.
-     */
-    @NonNull
-    public static Trace beforeCreateItem(@NonNull AbstractFolder<?> project,
```
jglick, line 116: https://github.com/jenkinsci/branch-api-plugin/blob/43d8bbd8b2c7ac526faf5d047af5b8e783cef681/src/main/java/jenkins/branch/OrganizationFolder.java#L1440 would need to be removed I suppose?

basil: > would need to be

https://github.com/jenkinsci/branch-api-plugin/pull/506

PR #480, opened 2025-04-04
Drop support for migrating from `branch-api` 1.x to 2.x

basil:
It has been 8 years since this migration, so anyone who is going to migrate already has by now. Drop support for this nonexistent use case to allow us to simplify the job loading code.

### Testing done

- Ran `branch-api` tests with these changes and verified that the only thing that broke were the migration tests (as expected). All other `branch-api` tests passed. Note that `integration.MigrationTest` tests that `OrganizationFolder`s can be loaded after a restart when their names were mangled.
- Testing in BOM in https://github.com/jenkinsci/bom/pull/4839

### Note

This PR breaks the migration tests in `branch-api`. https://github.com/jenkinsci/branch-api-plugin/pull/507 covers first deleting this tests. If accepted, that PR must be merged and released first (and adopted in BOM and any proprietary equivalents) before this PR can be merged and released.


src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -325,6 +292,7 @@ final ResultWithOptionalSave<File> ensureItemDirectory(@NonNull P parent, @NonNu
      * @param item the item.
      * @param legacyDirName the name of the directory that the item was loaded from.
      * @throws IOException if the ideal name could not be attached to the item.
+     * @deprecated removed without replacement
```
basil, line 295: This method is no longer used as of this PR. I could delete it if desired, but that would cause a compile-time error in `branch-api` once updating `branch-api` to a BOM version that contains this PR.

jglick: IIUC the overrides in `branch-api` can be deleted at any time, at which point we could also delete this method.

basil:
> Merged, release pending; I think that is enough to merge this, since we are only looking at PCT failures which should be resolved shortly?

Agreed.

jglick (approved):

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
basil: Yes, exactly.

jglick:
[CloudBees-internal issue](https://cloudbees.atlassian.net/browse/BEE-52209) FTR

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
jglick: ⇒ https://github.com/jenkinsci/branch-api-plugin/pull/510

PR #483, opened 2025-04-08
Check to see if `ChildNameGenerator.CHILD_NAME_FILE` can be removed

basil:


src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -377,19 +374,15 @@ public static <K, V extends TopLevelItem> Map<K, V> loadChildren(AbstractFolder<
...
+                if (name == null) {
+                    name = subdir.getName();
                 }
                 if (item instanceof AbstractItem) {
                     var abstractItem = (AbstractItem) item;
-                    if (itemFromDir != null && !legacyName.equals(name) && abstractItem.getDisplayNameOrNull() == null) {
+                    if (itemFromDir != null && !subdir.getName().equals(name) && abstractItem.getDisplayNameOrNull() == null) {
                         try (BulkChange ignored = new BulkChange(item)) {
-                            abstractItem.setDisplayName(childName);
+                            abstractItem.setDisplayName(name);
```
jglick, line 385: I am not sure this is right. Does it have test coverage? My understanding is that the original `childName` would have been a raw name, say `fix/xxx`, whereas `name` here is encoded like `fix%2Fxxx` (as opposed to `subdir.name` which would be mangled like `fix_2Fxxx.123ab`). We want the display name to be `fix/xxx` rather than its default value of `Item.name` which I think is `name` in this context.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -407,22 +400,13 @@ public static <K, V extends TopLevelItem> Map<K, V> loadChildren(AbstractFolder<
...
-        if (!computedName.equals(name)) {
-            throw new IllegalStateException("Computed name '" + computedName + "' does not match name read from file '" + name + "'");
+        String name = childNameGenerator().itemNameFromItem(this, item);
+        if (name == null) {
+            name = dir.getName();
         }
         return name;
     }
 
-    protected final I itemsPut(String name, I item) {
```
jglick, line 421: This method seems to be used by a CloudBees proprietary plugin.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
basil: It does not have test coverage. I have removed this code, as it was not necessary. Reasoning is as follows.

First, `branch-api` already sets the display name if needed:

```
src/main/java/jenkins/branch/MultiBranchProject.java
2049:                project.setDisplayName(displayName);
2119:                project.setDisplayName(getProjectDisplayName(project, rawName));

src/main/java/jenkins/branch/OrganizationFolder.java
1455:                            project.setDisplayName(projectName);
```

I verified empirically this was working as expected with organization folders, repositories, and branch names with special characters in the names.

Second, the logic was only being executed on reload, not initial load, and wasn't running at all for the past year and a half until https://github.com/jenkinsci/cloudbees-folder-plugin/pull/486 was released. And nobody complained. So this is further empirical evidence in favor of this being dead code.

Accordingly, I have deleted the dead code.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
basil: Restored with a deprecation warning asking people to inline this (now) trivial method.

PR #484, opened 2025-04-08
Prepare to eventually remove `ChildNameGenerator#CHILD_NAME_FILE`

basil:
`ChildNameGenerator#CHILD_NAME_FILE` introduces unnecessary complexity and has been at the scene of several concurrency issues, so we would like to eventually remove it. But it is not completely clear if it is still needed. I stepped through a few scenarios in the debugger in `cloudbees-folder`, `branch-api`, and `workflow-multibranch`, and I couldn't find a case where it was needed. But my confidence is low that I have fully understood the problem.

If there are no calls to `readItemName`, then it is pointless to write the item name and we can also remove all calls to `writeItemName`. So what would it take to remove all calls to `readItemName`? This PR computes the item name a new way, without relying on `readItemName`, and then calls `readItemName` to confirm that the results are the same. If the results are not the same, an `IllegalStateException` is thrown.

The idea is that if this change is delivered and no complaints occur, we can be reasonably sure that the new way of computing the item name does not introduce a regression. We can then delete the calls to `readItemName`, since we will be sure the replacement yields the same results. And as a result of deleting all calls to `readItemName`, we can also delete all calls to `writeItemName` as well, and the implementations of `readItemName` and `writeItemName` themselves.

If, on the other hand, a complaint does occur, then we will understand precisely the use case that needs to be tested before we can completely eliminate `CHILD_NAME_FILE`.

N/A

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
@@ -378,6 +378,13 @@ public abstract class AbstractFolder<I extends TopLevelItem> extends AbstractIte
                     }
                 }
                 name = childNameGenerator.writeItemName(parent, item, effectiveSubdir, childName);
+                String computedName = childNameGenerator.itemNameFromItem(parent, item);
+                if (computedName == null) {
+                    computedName = subdir.getName();
+                }
+                if (!computedName.equals(name)) {
+                    throw new IllegalStateException("Computed name '" + computedName + "' does not match name read from file '" + name + "'");
+                }
                 if (item instanceof AbstractItem) {
                     var abstractItem = (AbstractItem) item;
                     if (itemFromDir != null && !legacyName.equals(name) && abstractItem.getDisplayNameOrNull() == null) {
@@ -400,7 +407,15 @@ public abstract class AbstractFolder<I extends TopLevelItem> extends AbstractIte
 
     @Override
     public String getItemName(File dir, I item) {
-        return childNameGenerator().readItemName(dir);
+        String name = childNameGenerator().readItemName(dir);
+        String computedName = childNameGenerator().itemNameFromItem(this, item);
+        if (computedName == null) {
+            computedName = dir.getName();
+        }
+        if (!computedName.equals(name)) {
+            throw new IllegalStateException("Computed name '" + computedName + "' does not match name read from file '" + name + "'");
+        }
+        return name;
     }
 
     protected final I itemsPut(String name, I item) {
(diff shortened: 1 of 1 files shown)


Vlatombe:
So your proposal is to rely on `com.cloudbees.hudson.plugins.folder.ChildNameGenerator#itemNameFromItem`, instead of `ChildNameGenerator#CHILD_NAME_FILE`.
https://github.com/jenkinsci/cloudbees-folder-plugin/pull/82#discussion_r97208128 for motivation of the separate name file.

basil:
Yes, exactly. After migrating from Branch API 1.x (which is safe to assume as of https://github.com/jenkinsci/cloudbees-folder-plugin/pull/480), `CHILD_NAME_FILE` should contain the same information that is present in `ProjectNameProperty.class` and/or `BranchProjectFactory#branch` (or the folder name), which is what is used by `itemNameFromItem` to calculate the name.

src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -378,6 +378,13 @@ public static <K, V extends TopLevelItem> Map<K, V> loadChildren(AbstractFolder<
                     }
                 }
                 name = childNameGenerator.writeItemName(parent, item, effectiveSubdir, childName);
+                String computedName = childNameGenerator.itemNameFromItem(parent, item);
+                if (computedName == null) {
+                    computedName = subdir.getName();
+                }
+                if (!computedName.equals(name)) {
+                    throw new IllegalStateException("Computed name '" + computedName + "' does not match name read from file '" + name + "'");
```
jglick, line 386:
```suggestion
                    throw new IllegalStateException("Computed name '" + computedName + "' does not match name written to file '" + name + "'");
```
in this case I guess.

jglick (approved):
Looks like a good plan.

> my confidence is low that I have fully understood the problem

Since @stephenc has not touched Jenkins code in years I would guess that you currently understand the problem better than anyone else. I do not really follow what

> The point of CHILD_NAME_FILE is to allow an admin to figure out which folder is which.

was supposed to mean (that was over eight years ago): we do not expect admins to be inspecting `$JENKINS_HOME` folders as a rule, and if they do, they can learn to look up `BranchJobProperty` in `config.xml`.

> if this change is delivered and no complaints occur

suggests that this should be marked `developer` not `internal`, since you need a release, right? (Also, it is not “internal” if it might at least in theory throw an exception.)

jglick (approved):

PR #500, opened 2025-05-26
remove usages of commons-lang from core

mawinter69:
To be able to remove commons-lang from core all direct and indirect usages must be replaced with commons-lang3 or native Java

- use XS_DATETIME_FORMATTER2 instead of deprecated XS_DATETIME_FORMATTER from core
- replaced commons-lang with native Java

* remove usages of commons-lang from core

### Submitter checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,6 +71,10 @@
   <dependencies>
     <dependency>
       <groupId>io.jenkins.plugins</groupId>
+      <artifactId>commons-lang3-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.jenkins.plugins</groupId>
       <artifactId>ionicons-api</artifactId>
     </dependency>
     <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -71,10 +71,6 @@
   <dependencies>
     <dependency>
       <groupId>io.jenkins.plugins</groupId>
-      <artifactId>commons-lang3-api</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>io.jenkins.plugins</groupId>
       <artifactId>ionicons-api</artifactId>
     </dependency>
     <dependency>
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
@@ -46,7 +46,6 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import jenkins.model.TransientActionFactory;
-import org.apache.commons.lang3.StringUtils;
 
 /**
  * Provides a way for a {@link ComputedFolder} to break the association between the directory names on disk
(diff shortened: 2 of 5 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
```diff
@@ -293,7 +294,7 @@ public final String readItemName(@NonNull File directory) {
         File nameFile = new File(directory, CHILD_NAME_FILE);
         if (nameFile.isFile()) {
             try {
-                childName = StringUtils.defaultString(StringUtils.trimToNull(Files.readString(nameFile.toPath(), StandardCharsets.UTF_8)), directory.getName());
+                childName = Objects.toString(StringUtils.trimToNull(Files.readString(nameFile.toPath(), StandardCharsets.UTF_8)), directory.getName());
```
jglick, line 297: I really doubt we need https://github.com/apache/commons-lang/blob/7302e9a308f0a5d59f17f72b6f5326e76cd71e3a/src/main/java/org/apache/commons/lang3/StringUtils.java#L8834-L8837 here. Not even clear to me why the original code was call `trimToNull` at all. It certainly cannot be null, it should not need to be trimmed, and it should not have been empty. If the file exists, we should load its contents as is.

jglick: (And #483 deletes this code anyway.)

pom.xml
```diff
@@ -69,6 +69,10 @@
   </dependencyManagement>
 
   <dependencies>
+    <dependency>
+      <groupId>io.jenkins.plugins</groupId>
+      <artifactId>commons-lang3-api</artifactId>
+    </dependency>
```
jglick, line 75: Would prefer not to add this dep unless it is really offering some nontirivial function it would be onerous to do without.

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
mawinter69: Replaced this with the Util.fixEmptyAndTrim from core for now.

pom.xml
mawinter69: removed

src/main/java/com/cloudbees/hudson/plugins/folder/ChildNameGenerator.java
jglick: Thanks!

src/main/java/com/cloudbees/hudson/plugins/folder/computed/PeriodicFolderTrigger.java
```diff
@@ -107,19 +106,19 @@ private static long toIntervalMillis(String interval) {
...
-            interval = StringUtils.removeEnd(interval, "m");
+            interval = interval.substring(0, interval.length() - 1);
         } else if (interval.endsWith("d")) {
             units = TimeUnit.DAYS;
-            interval = StringUtils.removeEnd(interval, "d");
+            interval = interval.substring(0, interval.length() - 1);
         } else if (interval.endsWith("ms")) {
             units = TimeUnit.SECONDS;
-            interval = StringUtils.removeEnd(interval, "ms");
+            interval = interval.substring(0, interval.length() - 1);
```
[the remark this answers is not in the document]
viceice: - #501

PR #501, opened 2025-05-29
Mistakes in `PeriodicFolderTrigger` handling of ms-scale times

jglick:
https://github.com/jenkinsci/cloudbees-folder-plugin/pull/500#discussion_r2113189108 good catch @szjozsef though in fact this was using the wrong unit anyway, dating all the way back to https://github.com/jenkinsci/branch-api-plugin/commit/5e85b537e1ca543191967097fc25f00626f1e170, and it does not seem to matter since the minimum permitted interval is 1m and that is also the minimum offered in the GUI dropdown.


jtnord (approved):
Yikes... the ability to perform a task every 1ms....

PR #504, opened 2025-06-05
Use symbols for default icon and folder store

mawinter69:
While there is a mapping in core to map `icon-folder` to `symbol-folder`, this mapping is not applied in the command palette. As command palette only supports symbols it falls backs to `symbol-search`

Before:
image
image

After:
image
image

* Use symbols for default icon and folder store

### Submitter checklist


jtnord (approved):
Thanks!

PR #741, opened 2026-04-14
Move item dirs without `config.xml` aside; do not warn about them on every startup

jglick:
Amending #475 & #548. If `config.xml` is missing, printing an error every time Jenkins starts just means there will be log noise forever, unless and until an admin manually cleans it up. We may as well move such directories aside for possible troubleshooting and not produce any further log spam.

Doing this only when the child is at least a day old, to prevent improperly mangling child directories in weird NFS cases (say).

[CloudBees-internal reference](https://cloudbees.atlassian.net/browse/BEE-52162)

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
@@ -335,7 +335,6 @@ public abstract class AbstractFolder<I extends TopLevelItem> extends AbstractIte
      * @param <V>        the child type.
      * @return a map of the children keyed by the generated keys.
      */
-    // TODO replace with ItemGroupMixIn.loadChildren once baseline core has JENKINS-41222 merged
     public static <K, V extends TopLevelItem> Map<K, V> loadChildren(AbstractFolder<V> parent, File modulesDir,
                                                              Function<? super V, ? extends K> key) {
         return ExtensionList.lookupFirst(ChildLoader.class).loadChildren(parent, modulesDir, key);
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/ChildLoader.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/ChildLoader.java
@@ -89,16 +89,16 @@ public abstract class ChildLoader implements ExtensionPoint {
                 } else {
                     var subdirP = subdir.toPath();
                     try {
-                    if (Files.getLastModifiedTime(subdirP).toInstant().isBefore(Instant.now().minus(SystemProperties.getDuration(GRACE_PERIOD_PROP, Duration.ofDays(1))))) {
-                        var brokenChildrenDir = parent.getRootDir().toPath().resolve("broken-children");
-                        var brokenChildDir = brokenChildrenDir.resolve(subdirP.getFileName());
-                        if (!Files.exists(brokenChildDir)) {
-                            Files.createDirectories(brokenChildrenDir);
-                            Files.move(subdirP, brokenChildDir);
-                            LOGGER.warning(() -> xmlFile + " did not exist; moved " + subdirP + " to " + brokenChildDir + " for analysis");
-                            return null;
+                        if (Files.getLastModifiedTime(subdirP).toInstant().isBefore(Instant.now().minus(SystemProperties.getDuration(GRACE_PERIOD_PROP, Duration.ofDays(1))))) {
+                            var brokenChildrenDir = parent.getRootDir().toPath().resolve("broken-children");
+                            var brokenChildDir = brokenChildrenDir.resolve(subdirP.getFileName());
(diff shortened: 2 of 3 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/AbstractFolder.java
```diff
@@ -335,7 +335,6 @@ protected void initViews(List<View> views) throws IOException {
      * @param <V>        the child type.
      * @return a map of the children keyed by the generated keys.
      */
-    // TODO replace with ItemGroupMixIn.loadChildren once baseline core has JENKINS-41222 merged
```
jglick, line 338: https://github.com/jenkinsci/jenkins/issues/17797#issuecomment-4246330520

src/main/java/com/cloudbees/hudson/plugins/folder/ChildLoader.java
```diff
@@ -81,7 +87,23 @@ public <V extends TopLevelItem> V loadItem(AbstractFolder<V> parent, File subdir
                 if (xmlFile.exists()) {
                     item = (V) xmlFile.read();
                 } else {
-                    throw new FileNotFoundException("Could not find configuration file " + xmlFile.getFile());
+                    var subdirP = subdir.toPath();
+                    try {
+                    if (Files.getLastModifiedTime(subdirP).toInstant().isBefore(Instant.now().minus(SystemProperties.getDuration(GRACE_PERIOD_PROP, Duration.ofDays(1))))) {
```
jtnord: does this repo not have spotless enabled, the indentation is wrong here.

src/main/java/com/cloudbees/hudson/plugins/folder/ChildLoader.java
```diff
@@ -81,7 +87,23 @@ public <V extends TopLevelItem> V loadItem(AbstractFolder<V> parent, File subdir
...
+                        var brokenChildDir = brokenChildrenDir.resolve(subdirP.getFileName());
+                        if (!Files.exists(brokenChildDir)) {
+                            Files.createDirectories(brokenChildrenDir);
+                            Files.move(subdirP, brokenChildDir);
+                            LOGGER.warning(() -> xmlFile + " did not exist; moved " + subdirP + " to " + brokenChildDir + " for analysis");
+                            return null;
+                        }
+                    }
+                    } catch (IOException x) {
+                        LOGGER.log(Level.WARNING, null, x);
```
jtnord: please always leave a hint for administrators, just logging exceptions is unhelpful.

```suggestion
                        LOGGER.warning() -> () -> xmlFile + " did not exist, and an error occurred attempting to move the directory for analysis", x);
                        return null;
```

src/main/java/com/cloudbees/hudson/plugins/folder/ChildLoader.java
```diff
@@ -81,7 +87,23 @@ public <V extends TopLevelItem> V loadItem(AbstractFolder<V> parent, File subdir
...
+                    }
+                    LOGGER.warning(() -> xmlFile + " did not exist");
```
jtnord: a warning that a file does not exist does not tell an administrator what the impact there is.

```suggestion
                    LOGGER.warning(() -> "could not load the Item " + subdir.getFileName() + " as " + xmlFile + " did not exist. Future reloads or restarts will continue to attempt to load the item, until the directory is " + SystemProperties.getDuration(GRACE_PERIOD_PROP, Duration.ofDays(1)).toMinutes() + " minutes old, after which it will be moved for analysis");
```

(assumes that the granularity that anyone sane would set this to on a production instance would be minutes, otherwise it will log really user unfriendly Strings. This could actually calculate the time based on the file timestamp and the and then log that but that seems like overkill)

jtnord (dismissed):
Code looks fine (apart from a funky indent).
Error logs could be made more friednly to admins (these logs should be targeted to Jenkins administrators, not Jenkins devs)

src/main/java/com/cloudbees/hudson/plugins/folder/ChildLoader.java
jglick: > does this repo not have spotless enabled

Not currently, no.

src/main/java/com/cloudbees/hudson/plugins/folder/ChildLoader.java
jglick: The exception would include everything needed to dig deeper, but fine.

jtnord (approved):

src/main/java/com/cloudbees/hudson/plugins/folder/ChildLoader.java
jglick: ⇒ https://github.com/jenkinsci/cloudbees-folder-plugin/pull/743

PR #742, opened 2026-04-14
Log `Item.delete` errors from `ComputedFolder.updateChildren` directly

jglick:
Amends #150. We would like for errors thrown when deleting children to be printed to the Jenkins system log rather than to the indexing log. This would make it easier to diagnose issues after the fact, because indexing logs get rotated somewhat frequently, and anyways only Jenkins admins can resolve issues of this nature.

[CloudBees-internal issue](https://cloudbees.atlassian.net/browse/BEE-52160)


src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java
```diff
@@ -288,16 +287,12 @@ final void updateChildren(final TaskListener listener) throws IOException, Inter
                             existing.delete();
                             // super.onDeleted handles removal from items
                         } catch (IOException x) { // InterruptedException is still propagated
-                            deletingExceptions.add(x);
+                            LOGGER.log(Level.WARNING, "Failed to delete " + existing, x);
```
gbhat618, line 290: one question I had was — in case the children weren't deleted due to reason like - device busy - would the deletion be re-attempted at some point ?

but this PR doesn't change anything to retry behavior, so I guessing the reattempt will try to delete the orphaned children again. (but checking a bit)

gbhat618: answer was right in the `try` section above 😄 
> // super.onDeleted handles removal from items

so the child was not removed from the `items` - will be reattempted to delete in next cycle.

gbhat618 (approved):
lgtm. I don't have much experience with this plugin, but my understanding is - there is no deletion behavior change, it's just altering user reporting. As stated in the PR description - it will make the diagnosis easier 🎉

PR #747, opened 2026-07-25
[JENKINS-62218] Disable Default View select for read-only folder config viewers

Racknaraock:
# [JENKINS-62218] Disable Default View select for read-only folder config viewers

## Problem

`AbstractFolder/configure.jelly` sets a `readOnlyMode` variable (`<j:set var="readOnlyMode" value="${!it.hasPermission(it.CONFIGURE)}" />`) and already hides the Save/Apply buttons for a user without `Item/Configure`. Most fields inherit `readOnlyMode` automatically through Jenkins core's form tags, but the **Default View** selector is a hand-rolled `<select>` (not `f:select`), so it never inherited `readOnlyMode` at all — a read-only viewer could change the selection client-side (it just couldn't be saved, since Save/Apply are already correctly hidden).

This is part of the [JENKINS-12548](https://github.com/jenkinsci/jenkins/issues/12412) epic (read-only system configuration browsing / JEP-224).

**Note on prior scope:** earlier revisions of this PR also touched Health metrics (auto-expanding it out of `f:advanced` for read-only viewers) and added `capture="readOnlyMode"` to the Views Tab Bar / Icon `f:dropdownDescriptorSelector` calls. Per review discussion:
- The `capture="readOnlyMode"` gap is a generic Jenkins core bug (affects every `dropdownDescriptorSelector` consumer, e.g. `config-scm.jelly`'s `scmCheckoutStrategy`), so it was reverted here and pursued instead as [jenkinsci/jenkins#27167](https://github.com/jenkinsci/jenkins/pull/27167).
- Per feedback from @jtnord and @timja, advanced configuration sections should render the same regardless of `readOnlyMode` — if a section was previously collapsed behind `f:advanced`, it should stay that way in read-only mode too. That change (and its test) has been reverted.

This PR's scope is now just the Default View select fix.

## Changes

In `src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly`:

- **Default View**: add `disabled="${readOnlyMode ? 'true' : null}"` to the `<select name="primaryView">`, matching the idiom used by Jenkins core's own `f:dropdownList` (and by core's own `GlobalDefaultViewConfiguration/config.jelly`, which solves the identical "pick a default view" problem the same way).

### Default View — Viewer

Verified on a local `mvn hpi:run` instance with a Matrix-Authorization-Strategy `viewer` (`Item.READ`+`Item.EXTENDED_READ`+`View.READ`, no `Item.CONFIGURE`) against a folder with 2 views:

Before | After
:---:|:---:
before | after

Before the fix, the select had no `disabled` attribute and its value could be changed via a normal `change` event (confirmed via script: setting `.value` and dispatching `change` succeeded). After the fix, `disabled` is `true` and the control no longer accepts input.

Admin, for reference (no regression — still fully editable):

admin

## Testing

- Manual verification on a local `mvn hpi:run` instance, comparing an admin against a read-only viewer on the same folder's configure page (screenshots above), including live DOM/interactivity checks (`disabled` attribute state, attempted `change` events) for the Default View select.

## Checklist

- [x] Referenced the Jira issue (JENKINS-62218) in the commit messages
- [x] Interactively tested the change on a local Jenkins instance

* JENKINS-62218: The "Default View" selector on a folder's configuration page is now correctly disabled for read-only viewers.


src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
```diff
@@ -78,7 +78,7 @@ THE SOFTWARE.
               <j:if test="${it.folderViews.primaryModifiable and views.size()>1}">
                 <f:entry title="${%Default View}" field="defaultView">
                   <div class="jenkins-select">
-                    <select class="jenkins-select__input" name="primaryView">
```
jtnord, line 81: better to adapt this to `f:select`?

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
```diff
@@ -122,20 +122,32 @@ THE SOFTWARE.
...
         <f:section title="${%Health metrics}" icon="symbol-pulse-outline plugin-ionicons-api">
-          <f:advanced title="${%Health metrics}" align="left">
-            <f:block>
-              <f:hetero-list descriptors="${descriptor.getHealthMetricDescriptors()}"
-                             items="${it.healthMetrics}" name="healthMetrics"
-                             addCaption="${%Add metric}" deleteCaption="${%Remove}"
-                             oneEach="true" hasHeader="true" honorOrder="true"/>
-            </f:block>
-          </f:advanced>
+          <j:choose>
```
jtnord: if advanced blocks are not supposed to be expanded by default in read only mode then the change should go in Jenkins core so that all configuration items behave the same.

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
```diff
@@ -97,23 +97,23 @@ THE SOFTWARE.
               <j:if test="${it.folderViews.tabBarModifiable}">
                 <j:choose>
                   <j:when test="${descriptor.tabBarConfigurable}">
-                    <f:dropdownDescriptorSelector title="${%Views Tab Bar}" field="viewsTabBar"/>
+                    <f:dropdownDescriptorSelector title="${%Views Tab Bar}" field="viewsTabBar" capture="readOnlyMode"/>
```
jtnord: this looks strange.
if this is needed why is this not more prevelant elsewhere.  e.g. any project that defines an SCM -> https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/lib/hudson/project/config-scm.jelly#L45-L48 would appear to have issues?

jtnord (commented):
> Health metrics was wrapped in f:advanced, Jenkins core's collapsible "Advanced..." toggle. That toggle renders identically for every user regardless of permissions, so a read-only viewer has to click it just to discover whether any health metric is even configured.

Seems appropriate, not a bug.  Otherwise a config page can be full of loads of things you need to scroll through.
_If_ this is an issue (which I do not beleive it is) it should not be fixed here, but in Jenkins core so that the behaviour is the same for all consumers of `f:advanced` in all of the Jenkins UI.

> Views Tab Bar / Icon (f:dropdownDescriptorSelector) render their currently selected descriptor's config fragment inline, where readOnlyMode is visible normally — but any other descriptor's fragment is deferred via Jenkins core's l:renderOnDemand, which only exposes variables explicitly named in the tag's capture attribute. Neither call declared capture="readOnlyMode", so an extra config field from a third-party Icon/TabBar descriptor would come back fully editable to a read-only viewer once loaded.

Would appear to be a generic bug in Jenkins core and would affect all users of `dropdownDescriptorSelector` across the project so should be fixed in Jenkins core if so?

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
Racknaraock: Sooo... Choosing `f:select`  would require refactoring the control according to the AJAX-fill model; I decided that this change was better, since it is a pattern already used in the core 

https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/hudson/views/GlobalDefaultViewConfiguration/config.jelly#L26

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
Racknaraock: The strategy followed to implement these changes in other Epic tasks has been to implement them internally within the plugins—I understand this is to keep the components as generic primitives without any kind of permission context (such as read-only). Implementing the change in the core would mean altering all the components that use it, with the problems that could entail... 

I also understand your point—not including it in the core would lead to UX inconsistencies, but that’s more of a design decision that the project maintainers should make.

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
Racknaraock: It looks like so... I've reverted the change in the plugin. I'll open an issue in the core later to try to fix it there.

That was a good catch—thanks for letting me know.

Racknaraock: @jtnord I've oppened https://github.com/jenkinsci/jenkins/pull/27167

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
jtnord: > I also understand your point—not including it in the core would lead to UX inconsistencies, but that’s more of a design decision that the project maintainers should make.

FTR I think the current behaviour in this regard is correct. To me a readonly view should display the same as a read/write view with the exception of the disabled state of components, and the lack of save/apply buttons.

Racknaraock: @kohsuke  was the creator of [JENKINS-12548](https://github.com/jenkinsci/jenkins/issues/12412) ; perhaps he can provide some more information. I'm just trying to fix the reported issue. If the decision is to leave it as is, perhaps the issue should be closed.

jtnord: Kohsuke is no longer active in this project.  [JEP-224](https://github.com/jenkinsci/jep/tree/master/jep/224) does not explicitly state that the layout should or should not change only

> Many views have been modified to hide information that is not relevent to read only users.
> Jelly tags in the Jenkins core have been modified to display a read only state, i.e. text instead of inputs, icons for checkboxes

However I maintain that nothing should be changing advanced configuration to standard configuration just because it becomes read only, it is either always advanced, or never advanced), but will summon @timja as the author of the JEP for a second opinion.

timja: Advanced configuration does not change anything here, if it was previously advanced then it should still be under that as part of read only change.

(Advanced can be evaluated as part of separate UX review, it should only be used for non standard or uncommon configurations, as it will hide options and users may not see them)

Racknaraock: This means, this statement in the epic : _''every "advanced" sections should be expanded out of the box.''_ 

Is not longer valid? Just trying to confirm if we are only expecting the fix on the Default view select or there is something else to fix that im not noticing.

Once I get the confirmation I can push the revert of the advanced sections.

jtnord: > This means, this statement in the epic : ''every "advanced" sections should be expanded out of the box.''
>Is not longer valid? 

Correct.

> Just trying to confirm if we are only expecting the fix on the Default view select or there is something else to fix that im not noticing.

that should be all that is needed

Racknaraock:
@jtnord @timja Agreed on the Health metrics point — reverted in 853d481. The section now stays behind `f:advanced` regardless of `readOnlyMode`, and the associated test (`readOnlyViewerSeesExpandedHealthMetrics`) was removed. This PR's scope is now just the Default View select `disabled` fix.

src/main/resources/com/cloudbees/hudson/plugins/folder/AbstractFolder/configure.jelly
Racknaraock: Revert applied.

It would also be good to clarify the incorrect statement in the epic. I'll continue working on the remaining tasks in that epic, so I'll keep that in mind, but if any other contributors join, they might run into the same problem.

jtnord (approved):

PR #749, opened 2026-07-31
Add draft ItemConfigurator support for Folder

somiljain2006:
See  https://github.com/jenkinsci/configuration-as-code-plugin/pull/2881.

Add a draft `FolderItemConfigurator` demonstrating how the proposed `ItemConfigurator` extension point can be implemented by the CloudBees Folder plugin to create and configure `Folder` instances through JCasC. As a proof of concept, this implementation supports configuring a folder's name, display name, and description, illustrating that the API can be extended beyond freestyle and pipeline jobs.

Example configuration:

```
items:
  - folder:
      name: "team-alpha-jobs"
      displayName: "Team Alpha"
      description: "All CI/CD jobs for Team Alpha"
```

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/FolderItemConfigurator.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/FolderItemConfigurator.java
@@ -1,0 +1,52 @@
+package com.cloudbees.hudson.plugins.folder;
+
+import hudson.Extension;
+import io.jenkins.plugins.casc.ConfigurationContext;
+import io.jenkins.plugins.casc.ConfiguratorException;
+import io.jenkins.plugins.casc.ItemConfigurator;
+import io.jenkins.plugins.casc.model.CNode;
+import io.jenkins.plugins.casc.model.Mapping;
+import jenkins.model.Jenkins;
+import java.io.IOException;
+
+@Extension
+public class FolderItemConfigurator implements ItemConfigurator<Folder> {
+
+	@Override
+	public String getName() {
+		return "folder";
+	}
+
+	@Override
+	public Class<Folder> getTarget() {
+		return Folder.class;
+	}
+
+	@Override
+	public Folder configure(String name, CNode config, ConfigurationContext context) throws ConfiguratorException {
+		try {
+			Jenkins jenkins = Jenkins.get();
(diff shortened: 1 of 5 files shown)


src/main/java/com/cloudbees/hudson/plugins/folder/FolderItemConfigurator.java
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	@Override
+	public Folder configure(String name, CNode config, ConfigurationContext context) throws ConfiguratorException {
+		try {
+			Jenkins jenkins = Jenkins.get();
+			Folder folder = (Folder) jenkins.getItem(name);
+
+			if (folder == null) {
+				folder = jenkins.createProject(Folder.class, name);
```
timja, line 32: can you add tests as well to demonstrate how this works please

src/main/java/com/cloudbees/hudson/plugins/folder/FolderItemConfigurator.java
```diff
@@ -0,0 +1,52 @@
...
+			}
+
+			Mapping mapping = config.asMapping();
```
timja, line 35: I assume this doesn't work with credentials as yet? Credentials are the biggest pain point I'm aware of with folders and job dsl

somiljain2006: Not yet. At the moment, this configurator only supports creating/updating the folder itself, along with description and displayName. I haven't added support for configuring folder credentials yet.

somiljain2006: I implemented folder-scoped credentials support and verified that it works. However, adding cloudbees-folder and credentials as compile-time dependencies to JCasC core introduces architectural problems because those plugins already integrate with JCasC. Rather than introducing reflection or other workarounds, I think the correct home for the folder-specific configurators is the cloudbees-folder plugin itself.

timja: yes all plugin specific configurator's should be in the plugins themselves.

JCasC only provides direct implementations for Jenkins core features

